PR #90, opened 2025-11-23
Lesc

tnegrin:
Oops -- I should have made a branch to merge into LESC. The comments still hold but not sure if/where we want to merge

Added regression tests for LESC features.

@francisli -- I commented out the interactive tests because they kept prompting me to install Apple dev tools on my machine. The S3 tests were timing out for some reason.

Can you please take a look and we can talk about how to properly configure them on my machine

### DIFF
--- a/server/lib/s3.js
+++ b/server/lib/s3.js
@@ -165,6 +165,11 @@ function putObject (Key, filePath) {
   );
 }
 
+function reset () {
+  client = undefined;
+  signerClient = undefined;
+}
+
 export default {
   copyObject,
   createBucket,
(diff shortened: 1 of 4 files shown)


francisli (changes requested):
Hi Toby, I think I've fixed all the outstanding issues in the branch. Eslint passes, I've fixed the wrong imports causing Storybook to break, and I've restored the test helper file so that things run in CI properly.

However, you only committed tests for feedback, not holds/lesc routes. Can you check and push those up?

If you're using the Docker based setup, everything you run for the codebase needs to be run inside a running container because all the test related packages and tools are inside the container volume. If you're getting prompts to install Apple dev tools, to me that suggests you're running `npm test` outside the container in the host macOS.

PR #157, opened 2026-01-22
Link Facility Addresses to Maps URLs 

grahamplace:
Figma:
CleanShot 2026-01-21 at 21 44 29

Dev:
CleanShot 2026-01-21 at 21 43 43

You can test the navigator behavior in Chrome:
- Open devtools → three‑dot menu → More tools → Network conditions.
  - Uncheck Use browser default under User agent.
  - Pick an Android UA (or paste one).
  - Click the address link and confirm it uses geo:0,0?q=... in the anchor href.
image


francisli (approved):
Hmm, interesting, I wasn't aware of "geo:" scheme links...

Now if only Apple will allow us to change the default maps app and support it...!

PR #221, opened 2026-01-29
Update Styles and add error message

justin123yu:


client/src/lesc/components/PropertyForm.jsx
```diff
@@ -140,6 +141,10 @@ function PropertyForm () {
               <Chip.Group
                 key={form.key('property')}
                 {...form.getInputProps('property')}
+                onChange={(value) => {
```
francisli: Instead of overriding onChange on this form component, I'd suggest using the `onValuesChange` callback in `useForm` to watch for the property value to become LARGE. Then you can set a simpler state variable (i.e. `isLarge`) to use in the conditional for showing the message.

francisli (changes requested):
Some suggested changes...

Can you also update the seed file that adds the RESET center to include a fake dummy phone number?

Issue #177, opened 2026-01-24
User will see missing text prompts and have input constraints on CAD Number/Supervising Sergeant's Star Number fields in Incident Form


francisli:
<img width="699" height="216" alt="Image" src="https://github.com/user-attachments/assets/f03d8376-27dc-4d46-b37a-58ca5849d7da" />

https://www.figma.com/design/Q8kS4FJXh1TbbM8eDR7Z6Y/CareConnect?node-id=3154-37230&t=dNZncK0ssKdhSMVE-4

francisli:
Confirmed that missing text prompts appear. CAD Number is up to 10 alphanumeric, Star Number is up to 4 numeric digits

PR #233, opened 2026-01-30
Issue #177 --User will see missing text prompts and have input constraints on CAD Number/Supervising Sergeant's Star Number fields in Incident Form 

nutanp:
Issue #177 --User will see missing text prompts and have input constraints on CAD Number/Supervising Sergeant's Star Number fields in Incident Form 

[Added help text and min/max requirement for fields](https://github.com/sfbrigade/care-connect/commit/91ffda48bb7189b3cc082a283029b48762195ba6)

### DIFF
--- a/client/src/lesc/components/IncidentForm.jsx
+++ b/client/src/lesc/components/IncidentForm.jsx
@@ -2,7 +2,7 @@ import { useEffect, useRef, useState } from 'react';
 import { useNavigate, useSearchParams } from 'react-router';
 import { Head } from '@unhead/react';
 import { IconArrowLeft, IconCurrentLocationFilled } from '@tabler/icons-react';
-import { Box, Button, Container, Fieldset, Group, Loader, Stack, Text, TextInput, Title } from '@mantine/core';
+import { Button, Container, Fieldset, Group, Loader, Stack, Text, TextInput, Title } from '@mantine/core';
 import { useForm } from '@mantine/form';
 import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query';
 import { DateTime } from 'luxon';


client/src/lesc/components/IncidentForm.jsx
```diff
@@ -168,18 +168,21 @@ function IncidentForm () {
                 key={form.key('cadNumber')}
                 {...form.getInputProps('cadNumber')}
                 label={<>CAD number<span>*</span></>}
-                type='number'
-                inputMode='numeric'
+                minLength={2}
+                maxLength={10}
                 onFocus={() => setShowAddressForm(false)}
               />
+              <Text size='xs' c='#868E96'>CAD is provided by dispatch (MDT / radio).</Text>
```
francisli: `xs` is too small. Did you look at the result and compare against Figma?

When inspecting the designs, make sure you are selecting the exact component you want, and not one of its parent containers. In this case, if you select the actual Typography component, you see that it is specified as `md` text:

Image

francisli: Instead of using the hex color, #868E96, try looking up the matching Mantine theme color. That will make it easier later if we decide to globally tweak the color schemes.

https://mantine.dev/theming/colors/#default-colors

In this case you can use the theme color name `gray.6`.

client/src/lesc/components/IncidentForm.jsx
```diff
@@ -168,18 +168,21 @@ function IncidentForm () {
...
                 key={form.key('supervisorBadgeNumber')}
                 {...form.getInputProps('supervisorBadgeNumber')}
                 label={<>Supervising Sergeant’s Star Number<span>*</span></>}
-                type='number'
+                minLength={1}
+                maxLength={4}
                 inputMode='numeric'
                 onFocus={() => setShowAddressForm(false)}
               />
+              <Text size='xs' c='#868E96'>You can skip now and come back to fill out before custody transfer.</Text>
```
francisli: The text is also too far away from the bottom of the TextInput component, as they are being spaced out by the parent Stack component.

Try wrapping both the TextInput and the corresponding Text in a Box component (one right on top of each other) or if you need to put a little bit of space between them, in a Stack component with an appropriate gap size.

PR #247, opened 2026-02-02
Button to extend hold no longer shows after user arrives.

MapleSyrupy:


francisli (approved):
Since it's such a small change I'm going to make it myself...

PR #256, opened 2026-02-03
Toast Improvements

grahamplace:
Reworked our toast support to (a) match figma and (b) work better overall. Bunch of changes, apologies for the large PR but it was easier to just overhaul the whole thing in one go:
- Align colors w/ Figma, define semantic color constants in a new colors.js
- Match styles better (border radius, icons, etc)
- Allow passing a secondary message
- Allow user to dismiss toasts manually
- Add stories for toasts
- Uninstall unnecessary storybook onboarding add on
- Add a useMobile hook so we can more easily check if on mobile screen (and in this case center toasts if on mobile and top right if on desktop)

Desktop:
CleanShot 2026-02-02 at 21 14 36

Mobile:
CleanShot 2026-02-02 at 21 14 59


francisli (approved):
This is a case where the code was originally generated by an AI that didn't know how to utilize the Mantine component library...

I was able to reduce the markup and styles even more by using the Mantine Notification component as a base, and use Mantine theme color names.

PR #268, opened 2026-02-06
Ensuring incidents expire when last hold expires (fixes #263)

### LINKED ISSUE #263: User is left with an active/open Incident after the last hold has expired
If a user creates a hold (which creates a new Incident record), and lets that hold expire, the Incident remains active for the user. The expiration of holds should also check if there are any remaining active holds, and if not, mark the Incident as completed so that it no longer appears in the UI.

beaudrykock-moi:
### Updated `server/prisma/client.js`:

- Modified the `expire` function in the Prisma client extension.
- Added logic within the transaction to identify incidents associated with expiring holds.
- For each affecting incident, the system now checks if there are any remaining active deflections (holds).
- If no active holds remain for an incident, the incident record is automatically updated with a completedAt timestamp, effectively closing it.

### Created `server/plugins/scheduler.js`:

- Implemented a new Fastify plugin (`scheduler.js`) that runs a background task every 60 seconds.
- This task calls `fastify.prisma.deflection.expire()`, ensuring that hold expirations and subsequent incident closures happen automatically without requiring user interaction.

### Added unit tests in `server/test/prisma.test.js`:

Tests cover the following scenarios:

- Single Hold Expiry: Confirm that an incident with one hold is marked as completed when that hold expires.
- Partial Expiry: Confirm that an incident remains active (`completedAt: null`) if one hold expires but another remains active.
- Sequential Expiry: Confirm that an incident correctly transitions to completed only after the final active hold expires.

Closes #263


francisli (approved):
I'm approving this, but removing the "scheduler" plugin. It's more of a quick hack that works in a single-instance/development environment, but it is not appropriate as we start moving into production. If we ever need to scale up the number of front-end web instances, we don't want that to then scale up the number of schedulers all repeating the same work. I'll set up something more appropriate over the next week or so as I get things deployed to the stage/prod environments...

beaudrykock-moi:
Ah, good call. Makes perfect sense.

Issue #57, opened 2025-11-14
As a SD, I can receive a subject


beaudrykock-moi:
So that the system can track the custody transfer

AC:

- [ ] Scan or otherwise register the subject ID as presented by SFPD officer
- [ ] Receive confirmation that subject ID was scanned

coilin:
@beaudrykock-moi OMG my bad 👍

PR #269, opened 2026-02-06
[5] #57: As a SD, I can receive a subject 

grahamplace:
https://github.com/sfbrigade/care-connect/issues/57

  - Implement /api/deflections/:id/transfer endpoint that (in a transaction):
    - Transitions a subject from ONSITE_AWAITING_TRANSFER → AWAITING_INTAKE,
    - Updates bed counts (occupied +1, holds -1) 
    - Creates audit trail records 
  - Adds Custody page with simple tabbed view (In Custody / Released) listing subjects awaiting intake
  - Add full-screen QR scanner modal (ScanTransferCodeModal) for scanning transfer codes, with manual code entry fallback
  - Enhance QRScanner component with a fullScreen mode featuring viewfinder overlay, scan phase feedback
  (idle/pending/success/error)
  - Add `subjectStatus` filter to deflections list API
  - Allow custody-role users to query deflections by facility (see comment about requirements)

### DIFF
--- a/client/src/AppLayout.jsx
+++ b/client/src/AppLayout.jsx
@@ -5,7 +5,6 @@ import { matchPath, useLocation, useNavigate } from 'react-router';
 import { useQueryClient } from '@tanstack/react-query';
 
 import Header from './Header';
-import Navbar from './Navbar';
 
 import Api from './Api';
 import AppRoutes from './AppRoutes';
@@ -50,9 +49,6 @@ function AppLayout () {
           <Header opened={opened} close={close} toggle={toggle} logout={logout} />
         </AppShell.Header>
       )}
-      <AppShell.Navbar>
-        <Navbar close={close} logout={logout} />
-      </AppShell.Navbar>
       <AppShell.Main px={0}>
         <AppRoutes />
       </AppShell.Main>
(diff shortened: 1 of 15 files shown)


client/src/lesc/components/custody/Custody.jsx
```diff
@@ -0,0 +1,122 @@
...
+            fullWidth
+            value={tab}
+            onChange={setTab}
+            data={[
+              { label: 'In Custody', value: 'in-custody' },
+              { label: 'Released', value: 'released' },
+            ]}
+          />
+          {tab === 'in-custody' && (
+            <Stack gap='md'>
```
grahamplace: For this PR, this /custody page is just a bare bones page to list deflections, so that we can test the intake flow. 

https://github.com/sfbrigade/care-connect/pull/275 deals with the bulk of the real requirements for that page, so I would gloss over this file when reviewing this PR and focus on the QR scanner and transfer backend logic

server/routes/api/deflections/list.js
```diff
@@ -54,7 +55,11 @@ export default async function (fastify, opts) {
         where.status = status;
       }
 
-      if (!request.user.isAdmin) {
+      if (subjectStatus) {
+        where.subjectStatus = subjectStatus;
+      }
+
+      if (!request.user.isAdmin && !(request.user.isCustody && facilityId)) {
```
grahamplace: This needs a double check re: intended requirements. [Slack discussion](https://sfbrigade.slack.com/archives/C09PBES8WSU/p1770698512696439)

Current implementation is meant to allow SFSO users to see _all_ in custody subjects in their facility, hence why we _don't_ do the `where.createdById = request.user.id;` if user is custody user & query is facilityId filtered

PR #314, opened 2026-02-17
Add localStorage persistence for facility context

AnastasiaBelenkii:
## Summary
- Facility selection now persists across testing workflows
- Fixes FacilitySelector reappearing after adding new server routes

## Changes
- localStorage saves selected facility (dev only, production unaffected)
- SSR-safe with `typeof window` check
- Handles corrupted localStorage gracefully

## Context
Fixes issues where facility context was lost and FacilitySelector would reappear:
- After adding new server routes (reported by John)
- When using testing tools like agent-browser/playwright-cli to open specific routes
- Any scenario causing a full page reload would reset the in-memory facility state

Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>

### DIFF
--- a/client/src/AppLayout.jsx
+++ b/client/src/AppLayout.jsx
@@ -13,7 +13,7 @@ import { useFacilityContext } from './FacilityContext';
 function AppLayout () {
   const [opened, { close, toggle }] = useDisclosure();
   const navigate = useNavigate();
-  const { facility } = useFacilityContext();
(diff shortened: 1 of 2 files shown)


francisli (approved):
Tweaked the implementation a bit, and added some code so that the localStorage is cleared on explicit user logout... otherwise, there's no way to reset and get back to the facility selector.

PR #335, opened 2026-02-20
Facilitate voice dictation for Narrative field

saadabdali:
**Problem**: on the Deflections page, typing a long `Narrative` on a mobile device is tedious, and users might not be familiar with the voice dictation feature on their mobile device.

**In this PR:**
- On Android Mobile devices, we show a tappable microphone icon directly in the TextArea. User can tap the mic to begin speaking with transcription. The transcribed text is appended to any existing text in the Narrative field.
- On iOS Mobile, there's no way to hook into that functionality directly. So we show some hint text below the Text Area. User can tap the hint text for a modal that explains how to use the native dictation functionality from their iOS keyboard.
- On other devices (including desktop) we show a plain text area.
- This re-uses some existing device detection code from the `QRScanner`, so I extracted that code out to be shared.

## Android
Viewed in Chrome's Galaxy S8+ emulator:
image

Viewed and tested in a proper Android Studio Pixel 7 emulator:
image

## iOS
image
image

## Desktop (no change)
image


francisli (changes requested):
I'm not sure this experience is quite right yet, it doesn't give immediate feedback as the user is speaking, instead the recognition seems to just pop in at the end, I think because you've turned OFF interimResults and are waiting for isFinal to be true.

Per some discussion online, despite the documented intent of the settings, setting `continuous` to `false` and `interimResults` to `true` should work consistently across both Android and iOS. I did a quick test with these settings and printing events into the console shows that this works on both Android AND iOS.

We'd just need to manage the recognition text state separately from the form data state so that we can update it appropriately in the field and only set it in the form data once the recognition is ended by the user pressing the STOP button.

francisli (changes requested):
Per the screenshare on Zoom, testing on Google Pixel 8 Chrome shows all the intermediate results being appended together into the input box.

PR #336, opened 2026-02-20
Generates text in the arrest details narrative after observations add…

### LINKED ISSUE #330: As a SFPD Officer, I can have the person details narrative smartly pre-populated
So that I have to do minimal typing in the field.

AC:

- [ ] Pre-populate the narrative based on the selected observation chips in the person details section, using the structure below

"Officer encountered this individual at [place] on [date] at [time]. Officer observed the following behaviors: [insert behaviors from chips]. Officer concluded that a 647(f) RWS arrest was appropriate and transported the individual to RESET." (TBC with SFPD)

beaudrykock-moi:
### Summary
- Added a new narrative builder utility in `deflectionNarrative.js`.
- Updated `DeflectionForm.jsx` to auto-populate and auto-update the narrative text field based on incident details and selected behavioral observations.
- Added unit coverage at `deflectionNarrative.test.js`.

### Narrative template now used
Officer encountered this individual at [address] on [date] at [time].
Officer observed the following behaviors: [behaviors].
Officer concluded that a 647(f) RWS arrest and transport of the individual to RESET was appropriate.
(Note: this template may change with further feedback from SFPD stakeholders)

### Behavior changes
- Narrative is generated only after at least one behavioral observation is selected.
- Narrative updates automatically when selected observations change (including unselecting previously selected observations), and incident fields used by the template change.
- When no observations are selected, narrative is automatically cleared.
- Missing incident values in generated text use [ADD MANUALLY].
- Time formatting is normalized to avoid locale whitespace inconsistencies.

### Other chnages
-  Also updated subtitle text in this section to make it clearer where the text is coming from

### Test updates
Added tests for:
- full narrative generation with incident + behaviors,
- [ADD MANUALLY] placeholders for missing incident fields.

### Verified with:
`deflectionNarrative.test.js` (passing).

## Update 3/5/26
Commit refactors deflection narrative handling in the Field user “Add person details” flow to make narrative generation clearer, preserve user-added content across navigation, and keep persistence behavior consistent.

### Narrative UX redesign
- Replaced direct editable generated narrative with a two-part model:
  - **Auto-generated narrative** (based on selected observations and incident details)
  - **User additions** in a separate textarea
- Added a new narrative display section:
  - Title: **`647(f) narrative`**
  - Subtitle: **`This text will be inserted in the 647(f). Add to it using the form below.`**
- Updated second field label:
  - **`Add to narrative (optional)`**
- Removed previously introduced dirty-mode conflict prompt/controls.

### Persistence and payload behavior
- Kept server contract unchanged (`PATCH /api/deflections/:id` with `behavior`).
- Client now composes final persisted `behavior` as:
  - `generatedNarrative + "\n\n" + behaviorAdditions` (when both exist)
- Added unmount-time autosave flush so pending debounced edits are persisted on quick navigation (back/next/re-enter).

### Edit hydration fix
- Fixed a regression where manually added narrative was visible on detail view but disappeared when re-entering edit mode.
- Root cause: initialization path used payload-normalization (server-shape) instead of form-shape.
- Solution:
  - separated **form normalization** from **update payload building**
  - preserved `behaviorAdditions` in form state on load/reset.

### Placeholder language update
- Replaced missing-data token in generated narrative:
  - from **`[ADD MANUALLY]`**
  - to **`[DETAILS MISSING]`**
  
### Test Coverage Added/Updated

Added/updated unit tests to verify:
1. generated + additions composition order
2. additions-only persistence when no observations selected
3. regenerated base text retains additions
4. update payload shape (`behavior`, sorted `deflectionDetails`)
5. extraction of additions from persisted combined text
6. backward-compatible extraction when legacy text does not match generated prefix
7. placeholder behavior now uses `[DETAILS MISSING]`

## Update 3/5/26 4:04 pm
- Implemented narrative split for arrest details by adding behaviorAdditions (manual-only) in the API/data model and keeping behavior as the combined generated + manual text used for 647(f). 
- Updated DeflectionForm to hydrate the textarea from behaviorAdditions (no more reverse-extraction from combined text), and to continue composing/saving behavior for downstream views/PDFs. 
- Fixed a back-navigation autosave race where stale unmount saves could overwrite behavior with manual-only text, causing generated narrative to disappear on subsequent edits. 
- Added/updated client and API tests for payload persistence and null/trim behavior.

Closes #330

### DIFF
--- a/client/src/lesc/components/DeflectionForm.jsx
+++ b/client/src/lesc/components/DeflectionForm.jsx
@@ -290,7 +290,7 @@ function DeflectionForm () {
                 key={form.key('behaviorAdditions')}
                 autosize
                 {...behaviorAdditionsInputProps}
-                placeholder='E.g. “Subject was unable to stand without assistance and repeatedly stepped into traffic…”'
+                placeholder='E.g. “Person was unable to stand without assistance and repeatedly stepped into traffic…”'
               />
               <Button type='submit' mb='xl'>
                 {isNew ? 'Next: Personal property' : 'Save arrest details'}


francisli (changes requested):
I just tested this on my phone, and I'm unable to edit the Narrative in the textarea. Every time a character is entered, the textarea loses focus and the keyboard disappears.

Also, do we need to worry about officers losing custom edits if they change the selection of details?

francisli (changes requested):
There's still some weird behavior going on here, with regards to the generation/separation of the narrative text. To reproduce:

1. Hold a chair, finish the flow, including adding Arrest details and additional narrative.
2. View Details, observe that the fully composed narrative appears.
3. Edit Arrest, observe that the additional narrative is extracted separately into the field
4. Press the back arrow on the upper left to cancel/go back
5. Observe: only the additional narrative appears, not the fully composed narrative

Thinking: if we no longer allow the officer to EDIT the generated part of the narrative, then we should just generate it as needed for display/forms rather than try to append it with the user edited narrative. This would reduce/remove all the added complexity in the DeflectionForm.

I'm going to move on to other PRs for now, but will come back to this to work on it if it's still open when I loop back...

beaudrykock-moi:
Yeah, seeing that too, good catch. Agree that your solution is simpler. Will fix and push.

francisli (approved):
Ok, I think this is working well and looking good now...!

Issue #352, opened 2026-02-26
As a SFPD Officer, I should see manually transferred persons in my history


beaudrykock-moi:
Expected:
As a SFPD Officer, if the SD has entered a transfer code manually, and the code was accepted, the person should appear in my history tab. 

Actual:
The transferred person, if the transfer was manual via code, does not appear in my history even after refresh.

Repro:
1. Create a complete hold as SFPD user
2. Using a SFSO account, enter the transfer code manually
3. Observe the person does not move into history

PR #369, opened 2026-02-27
SFPD active list refresh and hold card handling (#352)

beaudrykock-moi:
### Summary
This PR overhauls SFPD Active holds / History behavior and fixes multiple UX/state issues in the holds workflow.

- Reworked SFPD tab semantics so holds move between tabs based on subjectStatus progression, not only hold status.
- Kept hold lifecycle/status model unchanged (ACTIVE remains valid after transfer), while still moving transferred records to History in UI.
- Added polling + non-blocking refresh behavior to prevent content flashing during background refetch.
- Fixed incident-card behavior so Active incident disappears when no active holds remain and appears at top of History above newly moved holds.
- Corrected History grouping so all holds are grouped under their actual incident (not only cancelled holds).
- Standardized incident subtitle presentation in History (always rendered, consistent dimmed styling, and 16px incident header labels).
- Added unit tests for the new tab/grouping/view-model logic.

### Key behavior changes
- Active tab now includes only holds in:
   - DETAINED, ONSITE_AWAITING_TRANSFER
- History tab now includes:
   - inactive holds (active=false)
   - active post-transfer holds: AWAITING_INTAKE, READY_FOR_INTAKE, ADMITTED, IN_CHAIR, RELEASED, EXITED
- Background polling updates lists without loader flash.
- Incident visibility:
   - Active: show incident only when Active has holds.
   - History: when all holds moved, show current incident above those moved holds.

### Technical changes
- Introduced shared view-model helpers for deterministic tab/group logic in `holdsViewModel.js`
- Refactored components to use shared helpers:
   - `Holds.jsx`
   - `HoldsActive.jsx`
   - `HoldsHistory.jsx`
   - `Incident.jsx`
- New unit test suite for tab/group/view-model behavior added in `holdsViewModel.test.js`, which covers:
   - subject-status filter constants
   - history merge + dedupe
   - active incident visibility rules
   - initial-load vs background-refresh states
   - current-incident pinning split
   - grouping by incident
   - ordering by latest hold activity
   - subtitle fallback generation


francisli (approved):
Ok, I really want to refactor how we're querying for deflections across all the different roles (sfpd, sfso, is), but this is working so I'm going to put that on the nice-to-have TODO list for now and merge this in so we can move forward...

PR #373, opened 2026-03-01
Issue 365: Label issue with IS scanning view

justin123yu:
Screenshot 2026-03-01 at 1 00 13 AM


justin123yu:
I noticed the design is not finalized so I will leave this as draft.

PR #406, opened 2026-03-11
Adding Admin UI to allow for editing deflection categories etc.

MapleSyrupy:


francisli (changes requested):
Looks good! Left some minor change request comments in the code.

One more thing to implement- for the admin list views, can you also show deleted records visually with a text strikethrough style. This will also require updating the list routes so that (only) an admin can request (i.e. through a query param) to include deleted records.

The reason why this is helpful is that, because we are "soft deleting" records, their ids will still be in the column, so they cannot be re-used. If someone were to try and re-create the same id/name, they would get an error but could not see why.

We could also add the ability for the admin to "undelete" a record.

francisli:
@MapleSyrupy I'm going to close this PR at this time- we're no longer using database tables to store these enum field values, but rather just Postgres ENUM types. As a result, these enum fields can only be updated in code, which will make it safer/easier to make sure that we have consistent values across environments. So we no longer need/support having an admin UI to edit/manage these values.

PR #437, opened 2026-03-12
UI adjustments and label changes

### LINKED ISSUE #413: Update "Mark complete" button title for SFSO user
Users were confused what this meant, what was being completed.

Correct to: "Safety checked"

### LINKED ISSUE #416: "Mark complete" button height does not match "View details" button
Actual: for SFSO user, "Mark complete" button is shorter than the "View details" button adjacent to it
Expected: button heights should match

Repro:
1. Scan someone in as SFSO user
2. Observe button heights on person card

Image

### LINKED ISSUE #417: Shorten overflow menu option titling
Existing overflow menu titles are too long.

Correct:

Record exit to jail >> Exit to jail
Start legal release >> Legal release

### LINKED ISSUE #424: Exit destination options should not use misleading "declined consent" language
In the Care user's exit destinations form: "Declined consent" might imply the person declined consent for a specific action, whereas this is really about the person not wanting to share.

Edit: "Declined consent"
To: "Did not share"

traceylum1:
Resolves #424 - changed all occurrences of 'declined consent' -> 'did not share'
Resolves #417  - shortened overflow menu titles
Resolves #416 - 'Mark complete' button matches 'View details' button size
Resolves #413 - change 'Marked complete' -> 'Safety checked'

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,7 +166,7 @@ CareConnect tooling lives inside this repository. After bringing up the Docker s
 
 1. **Geocode clinics**
 
-   Ensure `OPENROUTESERVICE_API_KEY` is defined in `server/.env`. Enter the server container and run the geocode command:
+   Ensure `AWS_LOCATION_ACCESS_KEY_ID` and `AWS_LOCATION_SECRET_ACCESS_KEY` are defined in `server/.env`. Enter the server container and run the geocode command:
 
    ```
    cd server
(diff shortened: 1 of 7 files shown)


francisli (changes requested):
@traceylum1 can you revert the changes to the _ids_ and _values_ so that we're only changing the display text?

Otherwise, we'll need to make some db schema changes and add a migration...

traceylum1:
> @traceylum1 can you revert the changes to the _ids_ and _values_ so that we're only changing the display text?
> 
> Otherwise, we'll need to make some db schema changes and add a migration...

@francisli Oops ok, I will revert those changes!

Would it make sense to change the _name_ values to "Did not share" while keeping _id_ and _value_ as "declined_consent"?

ie:
      id: 'declined_consent',
      name: 'Did not share',   <--------
      createdById: adminUser.id,
      updatedById: adminUser.id,

Or do we only want what is shown in the UI component to say "Did not share" and everything else remain the same?

PR #448, opened 2026-03-16
Hints to SFPD on missing incident and person details (#379)

### LINKED ISSUE #379: As a SFPD Officer, I can see easily which person details are incomplete
So that I'm not paging around trying to figure out what's missing

AC:

- [ ] Once SFPD Officer adds some details but not all of them, provide some kind of affordance that makes it clear which sections need finishing

Notes:
Could be as simple as changing the "Edit details" buttons in the person details view to "Finish details" for incomplete sections, but great to have design input.

FIGMA — https://www.figma.com/design/Q8kS4FJXh1TbbM8eDR7Z6Y/CareConnect?node-id=9391-87915&t=e8d36lfmcxdvsmp1-1

beaudrykock-moi:
## Summary

This PR improves how incomplete required details are surfaced across the SFPD hold workflow.

### Hold card person summary
- Updated the secondary line under the person name to show only known values.
- Added the fallback `Age and gender missing` when both DOB and sex are missing, per designs.
- Supported these display states, per designs:
  - DOB + sex: `38 y.o., Male`
  - DOB only: `38 y.o.`
  - Sex only: `Male`

### Required-field treatment in edit flows
Applied consistent incomplete-state styling across:
- Person details
- Narcotics details
- Incident details
- Arrest details (`647(f) narrative` behavior selection): this wasn't in designs b/c designs are a bit out of sync with more recent app changes, but I interpreted general design standards as best I could
- Personal property

For required text fields with missing values:
- Red outline
- Red placeholder matching the field label, e.g. `Enter last name`
- Helper text: `This field is required`

For required chip groups with no selection:
- Red-tinted chip styling
- Helper text: `Select one`

Additional arrest-details update:
- `647(f) narrative` now always shows a red required `*`
- Narrative guidance text turns red when no behaviors are selected and the incomplete state should be shown

### Delayed hint behavior
Incomplete hints no longer show immediately on first entry.

Behavior now:
- First entry into a new/empty form: no incomplete hints
- Auto-save while editing: does not trigger incomplete hints
- If the user leaves with some data entered but required fields still missing, hints are shown the next time they return

This gating was applied to:
- Person details
- Incident details
- Narcotics details
- Personal property
- Arrest details behavior selection

### Details view updates
In the hold details view:
- `Details incomplete` chip now uses the Figma error-light styling
- Incomplete states are shown under individual field labels as red `Incomplete`
- Section action buttons change from `Edit details` to `Finish details` when that section is incomplete
- Personal property also uses `Finish details` when the required property selection is still missing

### Incident creation and completeness fixes
- Starting a hold with no active incident now creates the incident immediately
- Newly created incidents now populate `arrestedAt` on creation - this was causing validation issues
- `Arrest date & time` was removed from the field-level incomplete hint treatment, since it is auto-populated in the creation flow and it doesn't make sense to flag it

### Save feedback
- Added the success toast when saving person details from the property step and returning to the home/holds view, per designs

### UI polish and bug fixes
- Fixed truncation of `Keep incident` in the cancel-incident modal
- Fixed layout alignment for `Arrest state` / `Arrest ZIP code` when required-field helper text is visible
- Adjusted native `datetime-local` styling to better match the intended entered-value appearance

## Tests
Added focused coverage for:
- Hold card person summary states
- Person details hint gating:
  - no hints on first visit
  - hints on re-entry after incomplete data
- Incident details first-visit hint suppression
- No-active-incident hold flow creating an incident with `arrestedAt`
- `isValidIncident` validation behavior

Also updated related component tests and test setup to resolve environment and import-path issues generated while adding this coverage.

Closes #379

### DIFF
--- a/client/src/components/Input.module.css
+++ b/client/src/components/Input.module.css
@@ -15,4 +15,5 @@
 
 .error {
   font-size: var(--mantine-font-size-sm);
+  margin-top: var(--mantine-spacing-xs);
 }
--- a/client/src/lesc/components/IncidentForm.jsx
+++ b/client/src/lesc/components/IncidentForm.jsx
@@ -29,7 +29,7 @@ import { useToast } from '@/components/ToastContext';
 import { useFacilityContext } from '@/FacilityContext';
 import { formatAddress } from '@/utils/format';
 import { getCurrentLocationAddress } from '@/utils/geocoding';
-import { isBlank, getMissingChipClassNames, getRequiredTextInputClassNames } from '@/utils/formStyles';
+import { isBlank, getRequiredTextInputClassNames } from '@/utils/formStyles';
 import classes from './IncidentForm.module.css';
 
 const requiredFieldError = 'This field is required';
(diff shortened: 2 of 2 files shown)


francisli:
@beaudrykock-moi 

>     * Ensure that the incident is created immediately when starting a hold, and not just when details are added and saved in the incident

Under what circumstances is this an issue? When creating the first hold, the Incident form is shown- if the user backs out of that page, then no Incident nor Deflection/Hold is created. If the user submits "Create incident & hold", then the Incident is created server-side along with the first Deflection/Hold.
 
>     * Ensure arrestedAt is populated on creation, otherwise it causes testing errors

This is always populated into the new Incident Details form when creating that first Incident and Hold as described above, so I think I still need to understand in which circumstances for the first issue.

>     * First entry into a new/empty form: no incompleteness hints

This is implemented and deployed.

>     * Auto-save while editing: do not trigger incompleteness hints

This is implemented an deployed.

>     * If the user leaves with some data entered but required fields still missing, hints are shown the next time they return

This is implemented and deployed.

PR #452, opened 2026-03-18
Add arrival confirmation modal

### LINKED ISSUE #318: As a SFPD Officer, I can't mistakenly tap "I've arrived"
So that I'm not marking arrival before I've actually arrived.

AC:

- [ ] Either a confirmation on tap, or an easy undo, per design preference

FIGMA — https://www.figma.com/design/Q8kS4FJXh1TbbM8eDR7Z6Y/CareConnect?node-id=1990-23526&t=e8d36lfmcxdvsmp1-1

jeanniefu:
Add arrival confirmation modal to prevent SFPD from accidentally tapping "I've arrived" button. Fixes #318.

Spec here: https://www.figma.com/design/Q8kS4FJXh1TbbM8eDR7Z6Y/CareConnect?node-id=1990-23526&t=e8d36lfmcxdvsmp1-1

Screenshot of local changes:
Screenshot 2026-03-17 at 3 21 13 PM

### DIFF
--- a/client/src/lesc/components/CancelHoldModal.jsx
+++ b/client/src/lesc/components/CancelHoldModal.jsx
@@ -1,9 +1,10 @@
 import { useState } from 'react';
-import { ActionIcon, Button, Group, List, Modal, Stack, Text, Title } from '@mantine/core';
+import { Button, Group, List, Modal, Stack, Text, Title } from '@mantine/core';
 import { IconX } from '@tabler/icons-react';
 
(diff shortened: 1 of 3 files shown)


jeanniefu:
A couple Qs:

1. Are there any tests I can update?
2. How closely do we need to match the Figma spec, and what is the best way to match it? I modeled this off of CancelHoldModal with some styling changes, but it doesn't look exactly the same as the Figma one.

PR #469, opened 2026-03-20
Switch to "Not in custody" tab when a person is legally released and scroll to that card

### LINKED ISSUE #423: When a person is legally released, move view focus to not in custody, still on site section
So that the user isn't confused at what just happened or where the person went.

AC:

- [ ] Switch to not in custody tab, scroll to on site section, show blue border highlight on recently moved card

jessemavi:
- Closes #423 

This PR does the following:
- Switches to the "Not in custody" tab when a person is legally released
- Highlights the card for the person that was moved
- Scrolls to the card if not already visible on the page

Code changes:
- Updated a couple of sessionStorage values in the releaseMutation `onSuccess` so state is updated in `Custody` and we move to the "Not in custody" tab and highlight the card. 
- Added a new `useEffect` to run when the tab has been changed and we can use the `highlightedId` to scroll to that card if needed. There is an existing `useEffect` that sets the `highlightedId` and does scrolling on the "In custody" tab but the `el` in that is null when we switch tabs as it doesn't look like the card is in the DOM at that point. The one I added runs after `highlightedId` has been changed and the DOM has already been rendered. Might be possible to consolidate these into one effect but could get complex with adding some waiting logic so having them be separate effects works well.

I didn't see a test file for  the `Custody` component so wasn't sure if we wanted to add a test for this behavior, mainly the tab switching after release.

Screen recording of legally releasing a person with and without scrolling:

https://github.com/user-attachments/assets/82b26c5c-05a0-4c1c-8761-e824feb079f6

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -34,6 +34,7 @@
     "jspdf": "^3.0.4",
     "leaflet": "^1.9.4",
     "luxon": "^3.7.2",
+    "mantine-form-zod-resolver": "^1.3.0",
     "pdf-lib": "^1.17.1",
     "posthog-js": "^1.290.0",
     "qrcode.react": "^4.2.0",
--- a/client/src/components/BooleanInput.jsx
+++ b/client/src/components/BooleanInput.jsx
@@ -2,7 +2,7 @@ import { Chip, Group, Input } from '@mantine/core';
 import { useUncontrolled } from '@mantine/hooks';
 import { useTranslation } from 'react-i18next';
 
-function BooleanInput ({ label, value, defaultValue, onChange, ...props }) {
+function BooleanInput ({ label, description, required, error, disabled, value, children, defaultValue, onChange, ...props }) {
   const { t } = useTranslation();
   const [_value, setValue] = useUncontrolled({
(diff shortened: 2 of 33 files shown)


client/src/lesc/components/custody/Custody.jsx
```diff
@@ -130,6 +130,22 @@ function Custody () {
     });
   }, [inCustodyDeflections, releasedDeflections]);
 
+  useEffect(() => {
+    if (tab !== 'released' || !highlightedId) return;
+    const el = document.getElementById(`custody-card-${highlightedId}`);
+    if (el) {
+      const rect = el.getBoundingClientRect();
```
beaudrykock-moi: `getBoundingClientRect` returns zero dimensions for anything hidden, etc., which could lead to the scroll being skipped. Only imagine this would happen with a slow render/some kind of race. Could fix with
`el.offsetParent !==null` to confirm it's laid out and visible? But it's a very minor point.

beaudrykock-moi (approved):
Couple minor points for some edge cases, but otherwise this looks all good to me.

PR #505, opened 2026-03-28
Changes person card behavior for easier SFPD viewing (#453)

### LINKED ISSUE #453: As a SFPD Officer, I can see all cards and QR codes until all holds are transferred
So that I don't lose a sense of where I am in the list of QR codes while scanning is in process.

AC:

- [ ] After SFSO scanning a SFPD QR code, it does not immediately disappear from the SFPD view
- [ ] SFPD view of the scanned QR code has some indication that the QR code has been scanned and should not be (cannot be) scanned again

beaudrykock-moi:
## Summary

This PR:

1. Updates the SFPD holds UI to support more visible transferred-hold behavior for multi-hold incidents.
2. Fixes a brittle Testcontainers setup in server tests.

### SFPD transferred holds behavior

This updates the SFPD user view when an SFSO user scans a transfer code.

- For incidents with more than one hold:
  - A transferred hold stays under `Current holds`
  - The status label changes to `Transferred`
  - The QR code is replaced with the checkerboard placeholder
- For incidents with only one hold:
  - The transferred hold moves to `History` as normal
  - It retains the `Transferred` label
- Once all holds in the incident have been scanned:
  - Any transferred holds still shown in `Current holds` move to `History`
  - They retain the `Transferred` label
- In the `History` tab:
  - Transferred cards do not show the checkerboard
  - Transferred cards do not show the `View Details` button
- Also. card padding was tightened to match the Figma history variant more closely (padding-md vs. padding-xl)

#### Implementation notes

- Added display-list logic to keep transferred holds in the active view only while the incident still has untransferred active holds
- Suppressed those same transferred holds from history until the incident is fully transferred
- Updated the hold card component to render different transferred states in active vs history views

### Testcontainers startup fix

This fixes an environment-sensitive failure in server integration tests.

**Problem (relying on AI here but this explanation seems reasonable:**
The server test helper previously forced Postgres and MinIO testcontainers onto the `care-connect` Docker network whenever `CI` was unset. In my environment, that prevented Testcontainers from observing published host ports correctly, causing a timeout startup failure. Codex recommended (and Claude agreed):

- Updates to `server/test/helper.js`
- Removing the default `withNetworkMode('care-connect')` behavior
- Adding custom network mode opt-in through `TESTCONTAINERS_NETWORK_MODE`

### Codex argument why

The test helper already connects to containers through `getHost()` and mapped ports, so a custom Docker network is not required for normal test execution.Making the custom network implicit created a brittle dependency on Docker topology that worked in some environments but failed in others.

### Opt-in behavior

If the custom network is still needed locally (personally I don't know when this would be the case, but in any case), tests can be run with:

```bash
TESTCONTAINERS_NETWORK_MODE=care-connect npm test
```
Closes #453


francisli (approved):
This is working as intended, so will approve and merge to deploy, but two pieces of feedback:

1) The checkerboard image replacing the QR code looks like something is broken. I'd suggest creating something similar to the "locked" state, except instead of a lock, a checkmark or something.

2) Tweaking the margins on the Hold card to match history layouts reflects on hold cards in the active states as well. There's going to be a lot of visual tweaks/changes coming later anyway, so I think this will be addressed later...

francisli:
3) Seeing everything suddenly disappear and move to history after the last transfer is completed feels a bit jarring after having the list now be stable for a while. Suggestion: what if we keep everything in Active until the SFPD user hits "I've left" and use that explicit user action to clear things into history?

PR #506, opened 2026-03-30
modify login, add mfa via email

### LINKED ISSUE #488: Add e-mail MFA on login
So that we can satisfy SSO expectations on security for launch.

AC:

- [x] On login, send challenge code to the user's e-mail on file.
- [x] Provide means to enter challenge code, check code and complete login

aaronhans:
Email-based MFA on login. After entering correct credentials, users receive a 6-digit verification code via email. Session is created only after code verification.

UI updates:

- Login.jsx now has 2 states, enter credentials and enter code.
- Use Mantine's PIN Input component for code entry (Got design approval for this and is in mocks now)
- Resend code button has countdown timer
- Create new account removed from login screen
- New email template with code

Database updates:

- Added 5 MFA fields to User model: mfaCode, mfaToken, mfaExpiresAt, mfaAttempts, mfaLastSentAt

Backend updates:

- POST /api/auth/login — now returns { mfaRequired, mfaToken, email } instead of creating a session
- POST /api/auth/verify-code — validates code, creates session on success
- POST /api/auth/resend-code — generates new code with 30-second cooldown

Test updates:

- 11 new server tests covering login MFA response, code verification, expiry, rate limiting, resend cooldown
- Updated authenticate test helper to handle MFA flow
- Fixed missing await in assets test (pre-existing bug exposed by MFA changes)
- Updated existing auth test for new MFA flow


francisli:
@aaronhans I was about to merge this in, but then I realized this would make it impossible to use our test accounts in the environments...

I was thinking maybe we should parameterize this so that we can enable/disable MFA on a per account basis. We can add a new column to User with a default TRUE for mfa, and then a superuser could disable it on a per-user basis...?

What do you think?

aaronhans:
Good plan. I added a change to add 
```
"mfaEnabled" BOOLEAN NOT NULL DEFAULT true;
```
and set the test accounts to disable MFA. 

We could add admin UI to flip the MFA value on accounts but idk if it should ever be turned off. Let me know if you want an admin UI and I'll add it.

francisli (approved):

Issue #380, opened 2026-03-04
As a SFSO or SFPD admin, I can do account management


beaudrykock-moi:
So that I can ensure access is tightly controlled.

AC:

- [ ] SFSO or SFPD admin can invite new accounts, and re-send account invites
- [ ] SFSO or SFPD admin can close an existing account and stop that user from accessing/doing anything in the app

Suggest a research spike in conjunction with broader auth conversations to figure out best technical approach here.

beaudrykock-moi:
@coilin after a conversation with the medical team, they would like to limit all "manage capacity" functions to medical admin users. This doesn't impact this story but I created a separate ticket here #503 - sharing for visibility.

PR #510, opened 2026-03-31
[2] Members list endpoint for org-scoped user management

grahamplace:
## Summary
- Adds `GET /api/organizations/:organizationId/members` endpoint
- Returns members grouped by status: invited, active, disabled
- Filters expired invites and deleted users
- Includes `isCurrentUser` flag for self-protection in UI
- Full test coverage for auth, org-scoping, grouping, and filtering

## Context
Part 2 of 5 in a stacked PR series for #380. First real feature — server-only endpoint for listing org members.

## Test plan
- [x] `npm test` passes (server)
- [x] Endpoint returns correct groupings
- [x] Non-org members can't access other orgs' member lists


server/routes/api/organizations/_organizationId/members/list.js
```diff
@@ -0,0 +1,88 @@
+import { User } from '#models/user.js';
+
+export default async function (fastify) {
```
grahamplace: The [Figma designs](https://www.figma.com/design/Q8kS4FJXh1TbbM8eDR7Z6Y/CareConnect?node-id=12232-87092&t=EovTOlSrtxjO0i0S-4) are totally new / different than the existing admin screens (which are more like a super admin for us to use) 

I think the simplest approach here is to just totally parallel path these new org-scoped admin flows, hence why I add these dedicated endpoints to the API 

Image

francisli (approved):

PR #562, opened 2026-04-08
Generalized scroll highlight for custody view

### LINKED ISSUE #483: Refactor and create a generalized scroll/highlight solution for Custody (and perhaps also the other) list
Per discussion in #469, in Custody.jsx there are now _four_ useEffect blocks between lines 102 - 162 doing pretty much the same time for different use cases.

Can we enumerate these and refactor into a single generalized solution for setting a scroll/highlight target for the list? And perhaps also make it applicable to the Holds (SFPD) and Care (IS) vlist views as well.

jessemavi:
- Resolves https://github.com/sfbrigade/care-connect/issues/483

There are currently 5 useEffects in `Custody` that all are related to scrolling to a card that is highlighted and has just been edited and were often running at the same time. This removes those useEffects and combines into one useEffect for scrolling.

Screen recording:

https://github.com/user-attachments/assets/fc0da920-7059-41d1-b0e5-a38bfa663e68


aaronhans (changes requested):
Love the code cleanup here!

I recommend grouping all the sessionStorage cleanup outside of the rAF to ensure they all happen:

if (highlightTarget) {
  setHighlightedId(highlightTarget);
}

// clean up all targets immediately
window.sessionStorage.removeItem('custodyHighlightTarget');
window.sessionStorage.removeItem('custodyScrollTarget');
window.sessionStorage.removeItem('custodyInCustodySectionTarget');
window.sessionStorage.removeItem('custodyReleasedSectionTarget');

jessemavi:
@aaronhans thanks for taking a look! This useEffect runs multiple times so if we clean up all targets immediately on the first time it runs then the `el` ends up being null when for example the tab changes and the effect is run again. So I think we need to keep the clean up for after we have found the element and scrolled if needed. Screenshots below show the difference.

Screenshot 2026-04-15 at 12 03 21 PM
Screenshot 2026-04-15 at 12 12 25 PM

francisli:
@jessemavi this looks good, but the scroll target isn't working for me in Firefox and Chrome (so far). Does it work for you?

Try resetting your database with the latest in the branch- the updated seed script creates a bunch more holds, so there's a longer page to test with. Scroll down, View Details, then go back- it should scroll back to the one you left but it doesn't for me.

francisli:
@jessemavi I just switched back to dev, and the scroll is broken for me there as well, so it's not necessarily your refactor, but something else going on...?

jessemavi:
@francisli looked into it and also see that the scroll is broken in `dev`.

Is the behavior below what we want for the user doing the following: 1. Scroll down, 2. View Details, 3. Go back?

This is on the `dev` branch so unrelated to my changes in this PR and not related to my previous PR. Looks like the element wasn't in the DOM at the time of `el.scrollIntoView` when you go back. Added a timeout in the `custodyScrollTarget` useEffect just to see if I could get it working along with some `scrollMarginTop` because of the fixed header. If this is the correct behavior I can add this to my PR into the single effect and maybe look into better solutions.


https://github.com/user-attachments/assets/c2ca3e47-04c2-4d6a-9052-c4535d81826e

francisli (approved):
Thanks @jessemavi, finally had a chance to get back to this after the launch last week...

It does seem that setTimeout is needed for the scroll to work, so I've incorporated that into the PR and am going to merge it in now with some other refactor/cleanup...

Issue #340, opened 2026-02-23
Anonymize a person record after 72 hours have elapsed from when the related hold/deflection is no longer active


beaudrykock-moi:
So that we can stay compliant with CJIS regulations.

AC:

- [ ] Anonymize a person record in AWS after 72 hours has elapsed from when the related hold/deflection is no longer active.
- [ ] Null out identifying columns: firstName, lastName, middleInitial, dateOfBirth, driverLicense, addressLine1, addressLine2, city, state, postalCode, localId, sex, race

PR #571, opened 2026-04-09
Anonymize subject records after 72 hours (#340)

grahamplace:
## Summary

- Adds `anonymizedAt` field and `/// @pii` annotations to the Subject model; `subject.anonymize()` Prisma Client extension nulls PII fields for eligible subjects and sets `anonymizedAt`
- A subject is eligible when **all** of these are true (see `server/prisma/client.js`):
  - `Subject.anonymizedAt IS NULL`
  - **no** `Deflection` for the subject is still in progress, where "in progress" = `status = ACTIVE` AND `subjectStatus NOT IN (EXITED, DEATH_IN_FACILITY, DEATH_IN_CUSTODY)`. This matters because the normal lifecycle never transitions `status` out of `ACTIVE` — terminal `subjectStatus` values are the real signal that a deflection is done
  - `COALESCE(MAX(Deflection.updatedAt), Subject.createdAt) ≤ cutoff`, where `cutoff = now - ANONYMIZE_CUTOFF_HOURS` (default 72h). Subjects with no deflections become eligible 72h after their own `createdAt`
  - Using `Deflection.updatedAt` (rather than `createdAt` or a terminal-transition timestamp) means the 72h clock resets whenever any deflection for the subject is touched, so post-handoff edits keep PII available until activity actually stops
- Registers an hourly pg-boss cron job to run the anonymization; gated behind `DISABLE_ANONYMIZE_SUBJECTS` env var
- `ANONYMIZE_CUTOFF_HOURS` env var allows overriding the 72-hour default for testing
- Anonymization also clears S3-backed PII assets (generated 647f PDFs and property photos): uses the model's `setAsset('file', null)` helper so S3 deletion goes through canonical tooling, and the DB row is kept with `file` nulled in place — mirrors the anonymize-in-place pattern used for Subject
- Makes `DeflectionDocument.file` and `PropertyPhoto.file` nullable (migration) to support in-place nulling
- `/// @pii-asset` schema annotations on `DeflectionDocument.file` and `PropertyPhoto.file` flag S3 references containing PII
- Deflection list and detail API endpoints filter out anonymized subjects; `list` uses a Prisma `isNot` relation filter so null-subject deflections still pass through
- `PUT /api/deflections/:id/subject` returns 404 when the subject is already anonymized (defense-in-depth against re-populating PII post-anonymization)
- Seed script now creates test documents and property photos in MinIO, with cleanup on re-seed; also stops writing `status = COMPLETED` on seeded deflections (a value real code never produces)
- Schema validation test ensures `@pii` annotations stay in sync with the `PII_FIELDS` code constant

## Test plan
- [ ] `node --test test/jobs/anonymizeSubjects.test.js` — covers: eligible terminal deflection past cutoff, active/in-progress deflection skipped, ACTIVE-but-terminal-subjectStatus anonymized, within-cutoff window skipped, already-anonymized preserved, multi-deflection `MAX(updatedAt)` timing, subject with no deflections past cutoff, recently-created subject with no deflections skipped, S3 deletion + DB row preservation
- [ ] `node --test test/models/subject.test.js` — validates `@pii` schema annotations match `PII_FIELDS`
- [ ] `node --test test/routes/api/deflections.test.js` — covers list filtering and the new PUT-subject 404 for anonymized deflections
- [ ] Verify job no-ops when `DISABLE_ANONYMIZE_SUBJECTS=true`
- [ ] Manual: set `ANONYMIZE_CUTOFF_HOURS=0.016` and `DISABLE_ANONYMIZE_SUBJECTS=false`, trigger job ad-hoc, confirm in Prisma Studio that `Subject.anonymizedAt` is set with PII fields null, `DeflectionDocument` and `PropertyPhoto` rows are kept with `file = null`, and that the corresponding objects are gone from MinIO


server/prisma/client.js
```diff
@@ -132,6 +135,56 @@ const prisma = new PrismaClient({
         }));
       }
     },
+    subject: {
+      async anonymize (now = new Date()) {
+        const parsed = parseFloat(process.env.ANONYMIZE_CUTOFF_HOURS);
+        const hours = Number.isFinite(parsed) ? parsed : 72;
+        const cutoff = new Date(now.getTime() - hours * 60 * 60 * 1000);
+        const eligible = await prisma.$queryRaw`
```
grahamplace, line 122: This is the jankiest part of this PR and the most like to be brittle in the future. There's just not a super clean "terminal" timestamp to base the 72hrs calculation off of.

An alternative is we just decide the anonymization logic can be much simpler, something like "anonymize everything 72 hrs after creation", and assume that 72 hrs is enough time for someone to be fully out of the facility, but I don't know if that's a reasonable assumption. We don't want to wipe data of someone currently in the facility. Any thoughts on this @beaudrykock-moi ?

server/prisma/ERD.md
```diff
@@ -540,7 +541,7 @@ DEATH_IN_CUSTODY DEATH_IN_CUSTODY
     String id "🗝️"
     Int deflectionId 
     String formId 
-    String file 
+    String file "❓"
```
grahamplace, line 560: Need to make these nullable so that we can delete the field when anonymizing. We keep the record for referential integrity, but delete the file from s3 and this file pointer from DB

server/prisma/schema.prisma
```diff
@@ -489,23 +489,39 @@ model FacilityEligibility {
 
 model Subject {
   id                           String                  @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
+  /// @pii
```
grahamplace, line 499: These annotations aren't the SoT at runtime for what to clear in the anonymization job, but we do have a test that ensures that set matches these annotations

I want to annotate the schema so future editors know to consider PII concerns

server/lib/jobQueue/queueConfig.js
```diff
@@ -41,5 +42,11 @@ const queues = [
     handler: async ([job]) => formsEmail(job.data),
     deadLetterData: (data) => ({ deflectionId: data?.deflectionId }),
   },
+  {
+    name: QUEUE_ANONYMIZE_SUBJECTS,
+    options: { retryLimit: 1 },
+    handler: async ([job]) => anonymizeSubjects(job.data),
+    cron: '0 * * * *',
```
grahamplace, line 49: All anonymization handling (DB, s3) is run in this hourly pg-boss cron job

server/prisma/client.js
beaudrykock-moi: What if we did 96 hours from creation? That gives 24 hours for the maximum stay time at RESET, and 72 hours after that. I am assured that 24 hours really is the stay limit, but allowing for some slippage, I think 72 hours after that is as safe as we can reasonably be.

francisli (approved):

PR #572, opened 2026-04-10
[Do not merge] Prototype: Streamlined subject details flow

saadabdali:
4/29: [preview removed]

### DIFF
--- a/client/src/Api.js
+++ b/client/src/Api.js
@@ -510,6 +510,17 @@ const Api = {
       },
     },
   },
+  ai: {
+    parseId (image, mediaType) {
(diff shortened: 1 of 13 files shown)


saadabdali:
Notes:
- use only AWS services if possible (for the vision model, language model, etc)
- record raw voice narration
- probably need checkbox at the end where officer affirms they've reviewed everything

saadabdali:
Wondering if we can also drop the 'person volunteered for RESET' field. If narration is easy, we can likely rely on officer to capture that in their written description

saadabdali:
closing, because this functionality was written and merged into the app in other PRs.

PR #625, opened 2026-04-10
Fix/534 jail route in chair

### LINKED ISSUE #534: Allow exit to jail route from IN_CHAIR status
It appears that `exit-to-jail` is not available for an `IN_CHAIR` status.

Exiting to jail can happen at any time after custody transfer.

AC:

- [ ] Ensure that exit to jail from IN_CHAIR status is possible and correctly persisted on the backend

beaudrykock-moi:
## Summary

Fix `exit-to-jail` so it works for `IN_CHAIR` records after custody transfer and persists the expected downstream state for custody and reporting views.

## Changes

- allow `IN_CHAIR` in `POST /api/deflections/:id/exit-to-jail`
- persist jail exit state needed by the custody UI:
  - `subjectStatus: EXITED`
  - `exitDestinationId: 'jail'`
  - `releasedAt: null`
- persist `refusalReasonId` for jail exits
- mark associated property as returned on jail exit:
  - `propertyReturned: true`
  - `propertyReturnedAt`
  - `propertyReturnedById`
- fix bed accounting so:
  - pre-intake statuses release a hold
  - `IN_CHAIR` releases an occupied chair
- add regression coverage for `IN_CHAIR -> exit-to-jail`

## Why

The frontend custody UI was already wired to offer “Record exit to jail” for `IN_CHAIR` records, but the backend rejected that path with `409`. After merging with `dev`, the route also needed to preserve the newer `refusalReasonId` behavior while keeping the property-return persistence added for this fix.

Closes #534


aaronhans-sf (approved):
- Code changes look good
- Tested locally, used exit to jail from details screen for someone who was "in chair" status and then they showed up as not in custody under the "Transferred to jail" section
- Data after this looks good:

Status: EXITED with exitDestinationId: jail
Release: releasedAt is null (no legal release, direct jail exit)
Property: marked as returned with timestamp and user
Audit trail: EXITED update with jail destination and property returned

francisli (approved):

PR #631, opened 2026-04-13
store a hash of field values for 647 

### LINKED ISSUE #492: As a IS user, when I make person detail changes, this updates the 647(f)
So that any changes to the 647f are safely filed with SFPD.

AC:

- [ ] When IS makes changes to a user profile for someone in chair, the 647f should be re-generated, and filed when that person is released

aaronhans:
We don't want to refie 647 if there are no changes.

Right now we attach 647 to the release email always but we don't need to do that if there were no changes to it because it gets filed when the person gets checked in.

This PR adds a hash value to deflectionDocument table for the 647f form only.

When we are attaching files we check against prior hash if we have one already and do not attach form to email if hash matches.


francisli (approved):
Any reason not to do the same check and skip for the 849b and CoR...?

Issue #611, opened 2026-04-10
Change title of "Not in custody" tab


beaudrykock-moi:
To avoid confusion over what "in custody" means

AC:

- [ ] Change "Not in custody" tab label to "Legally released"

saadabdali:
Presumably this should affect both the SFSO view and the Care view?

beaudrykock-moi:
Yep - both SFSO and Care views. Thanks for clarifying!

PR #635, opened 2026-04-14
Update 'not in custody' to 'legally released'

saadabdali:
Addresses #611. 

This PR changes:

- The text label in the segmented control seen by CARE and SFSO users ("Legally released" rather than "Not in custody")
- Changes related user-facing text in a few toast notifications
- Updates `In Custody` to `In custody` (this was inconsistent across the CARE and SFSO views; this PR makes it consistent)

This PR **doesn't** change the underlying tab names used for navigation. (So the page still recognizes `custody?tab=not-in-custody` rather than `custody?tab=legally-released`. But we changed the user-facing copy in the UI.

image

image

### DIFF
--- a/client/src/lesc/components/care/Care.jsx
+++ b/client/src/lesc/components/care/Care.jsx
@@ -182,7 +182,7 @@ function Care () {
             onChange={setTab}
             data={[
               { label: 'In custody', value: 'in-custody' },
-              { label: 'Not in custody', value: 'not-in-custody' },
+              { label: 'Legally released', value: 'not-in-custody' },
             ]}
           />
           <FacilityStatusBanner />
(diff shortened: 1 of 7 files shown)


francisli (approved):

beaudrykock-moi (commented):
Looks great, but I wonder if we should standardize on the use of quotes. e.g. in `LegalReleaseQuestions.jsx` and `CareFlow` there are no quotes (i.e. = Legally released) but in `CustodyDetailContent` there are double quotes (i.e., = "Legally released") while `ExitToJailModal.jsx` has single quotes (i.e., = 'Legally released'). Maybe use double quotes for all these mentions?

saadabdali:
> Looks great, but I wonder if we should standardize on the use of quotes. e.g. in `LegalReleaseQuestions.jsx` and `CareFlow` there are no quotes (i.e. = Legally released) but in `CustodyDetailContent` there are double quotes (i.e., = "Legally released") while `ExitToJailModal.jsx` has single quotes (i.e., = 'Legally released'). Maybe use double quotes for all these mentions?

Sorry, it looks like I introduced the inconsistency here through inattention. I can do the fix you suggest.

PR #638, opened 2026-04-14
Allows manual SFPD unit entry and profile editing (#603)

### LINKED ISSUE #603: In the "what unit are you assigned today" view, allow manual entry of unit
So that SFPD can just type the full unit number even if it doesn't show in the list

AC:

- [ ] Add subtitle: "If unit number does not appear in list, type full unit number and confirm"
- [ ] Change placeholder: "Type unit name"
- [ ] Enable "Confirm unit" as long as at least 3 characters are entered
- [ ] Persist a manually entered unit on backend

beaudrykock-moi:
## Summary

Updated SFPD/SFSO unit entry so officers can type and save units that are not already in the system.

### Login unit selection behavior changes

- Added helper subtitle above first-login unit entry:
  - `If unit number does not appear in list, just type and confirm`
- Changed first-login unit placeholder to:
  - `Type unit name`
- Enabled `Confirm unit` once at least 3 characters are entered
- Updated first-login unit submission to:
  - match existing units case-insensitively
  - create a new backend `Unit` record when no match exists
  - save the resolved `unitId` onto the user

### Profile edit behavior changes

- Updated profile unit editing to match first-login behavior
- Replaced the unit dropdown with typed autocomplete
- Allowed users to save a manually entered unit even if it is not already in the known unit list
- Limited profile PATCH payloads to editable fields only, fixing a `403 Forbidden` issue caused by sending extra protected fields

### Backend

- Extended user update schema to accept `unitName`
- Added backend resolution logic to:
  - reuse an existing unit in the same organization when names match
  - create a new organization-scoped unit when needed
  - persist that unit association on the user

### Tests / verification

- Added backend coverage for:
  - creating and assigning a manually entered unit
  - reusing an existing unit on manual entry
- Lint passed for the changed client/server files
- Full server route test execution could not be completed locally in this environment because the test suite requires a working container runtime for `testcontainers`

### Notes

- Ideally, SFPD would give us a complete list of units. They have not done so, and may not even have a complete list of units that can be updated often enough. While allowing manual entry is not ideal and could lead to mistaken or incorrect units in the dataset, it may be the best we can do for now under the circumstances.
- Note that this applies to the SFSO as well, in case we run into the same problem, but they appear to have smaller numbers of units so less likely to come up.

Closes #603

### DIFF
--- a/client/src/UnitSelector.jsx
+++ b/client/src/UnitSelector.jsx
@@ -1,5 +1,5 @@
 import { useState } from 'react';
-import { Box, Button, Container, Stack, Title, Autocomplete, Loader } from '@mantine/core';
+import { Box, Button, Container, Stack, Title, Autocomplete, Loader, Text } from '@mantine/core';
 import { useQuery, useQueryClient, useMutation } from '@tanstack/react-query';
 import { useLocation, useNavigate } from 'react-router';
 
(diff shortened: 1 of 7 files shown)


aaronhans (approved):
Update works fine for me locally, no issues noticed in code changes

FYI there is identical slugify function https://github.com/sfbrigade/care-connect/blob/4a5d4faeb5d5266a789dd24b91f0e42dae437bbe/server/prisma/seeds/units.js#L67

beaudrykock-moi:
> Update works fine for me locally, no issues noticed in code changes
> 
> FYI there is identical slugify function
> 
> https://github.com/sfbrigade/care-connect/blob/4a5d4faeb5d5266a789dd24b91f0e42dae437bbe/server/prisma/seeds/units.js#L67

Ah good catch. I pulled the slugify dupe into a utility function.

PR #642, opened 2026-04-15
Adds DOB 4 digit validation and warnings (#639)

### LINKED ISSUE #639: Date of Birth doesn't warn or retain when 2 digits year entered
An arresting officer who:

- Creates an incident
- Goes back to add details
- Enters a date of birth with only 2 digits for the year like 11/11/01
- Gets no warning about invalid date
- The DOB is not retained

tagging @beaudrykock-moi to prioritize

beaudrykock-moi:
## Summary

Fix DOB validation in the person details flow so 2-digit years are handled clearly and without interrupting normal typing.

## Details

Users now get a clear explanation when DOB is missing a 4-digit year, cannot complete person details until DOB is valid, and can type a valid DOB naturally without premature error states.

- Enforced a 4-digit year for DOB before person details can be completed
- Added a specific DOB validation message for incomplete years like `11/11/01`
- Prevented autosave from silently clearing DOB when the user has only entered a partial date
- Adjusted the UX so partial DOB input does not show an error while the user is still typing
- Kept inline validation for complete-but-invalid DOB values
- Showed the 4-digit-year error on blur or submit when the field is incomplete
- Added focused unit tests for DOB validation behavior

## Why

Previously, entering a DOB such as `11/11/01` produced no clear warning and the value was not retained, which made the form feel broken and allowed users to miss why person details were still incomplete.

This change preserves the existing app patterns for field-level errors and workflow gating, while avoiding the jerky experience of showing an error during normal partial entry.

Closes #639


client/src/lesc/components/SubjectForm.jsx
```diff
@@ -130,6 +122,24 @@ function SubjectForm () {
...
+      ...normalized,
+      drugType: normalized.drugUseEvidence ? normalized.drugType ?? null : null,
+    };
+
+    const dobValue = String(dobString ?? '').trim();
+    if (!dobValue) {
+      payload.dateOfBirth = null;
+    } else if (parsedDob.isValid) {
+      payload.dateOfBirth = parsedDob.toISO();
+    }
```
aaronhans: Think we need an else branch here to set it to null if it isn't valid

Also see if we can prevent posting to server if we are missing a valid date

aaronhans (changes requested):

francisli (changes requested):
While testing, I notice that this implementation breaks the ability to save/submit the form while incomplete.

It also seems significantly overcomplicated- can we not just customize the Zod validator on dateOfBirth? If this is still open by the time I finish my pass through PRs, I'll swing back to see if I can find a simpler fix...

beaudrykock-moi:
Valid point. I’ll look into it

From: Francis Li ***@***.***>
Date: Friday, April 24, 2026 at 3:55 PM
To: sfbrigade/care-connect ***@***.***>
Cc: Kock, Beaudry (ADM) ***@***.***>, Author ***@***.***>
Subject: Re: [sfbrigade/care-connect] Adds DOB 4 digit validation and warnings (#639) (PR #642)


This message is from outside the City email system. Do not open links or attachments from untrusted sources.



@francisli requested changes on this pull request.

While testing, I notice that this implementation breaks the ability to save/submit the form while incomplete.

It also seems significantly overcomplicated- can we not just customize the Zod validator on dateOfBirth? If this is still open by the time I finish my pass through PRs, I'll swing back to see if I can find a simpler fix...

aaronhans:
Good point @francisli I refactored this

beaudrykock-moi:
@francisli @aaronhans just a reminder to take a look at the changes and see if they pass muster!

francisli (approved):

PR #644, opened 2026-04-15
Adds fallback incident completion worker (#585)

### LINKED ISSUE #585: Add fallback incident completion time
So that if the SFPD officer forgets to tap "I've arrived", we still have data in the backend to use for departure.

AC:

- [ ] If more than 30 minutes elapse after the last hold in the incident was transferred or canceled, automatically close the incident and mark the user as "I've left".
- [ ] In this instance, use the custody transfer time of the last hold as the "I've left" time

beaudrykock-moi:
## Summary

Adds an automatic incident close-out path for the case where an SFPD officer forgets to tap "I've left" after the last hold has already been transferred or canceled.

When an arrived incident has no remaining pre-transfer holds, the existing minute-based background reconciliation now auto-closes it after the configured delay (set to 30 minutes). 

## Details

- Added `incident.autoCloseAfterFinalHold()` in the Prisma client extension.
- Hooked the new incident reconciliation into the existing `expireHolds` background job, which already runs every minute.
- Auto-close logic now:
  - checks for incidents that were already marked as arrived
  - confirms there are no remaining active pre-transfer holds
  - waits until the final hold transfer/cancel event is older than the configured threshold
  - marks open `IncidentOfficer` records as left
  - sets `Incident.leftAt`
  - sets `Incident.completedAt`
- For the "I've left" timestamp:
  - prefers the most recent `transferredAt`
  - falls back to `cancelledAt` if the final hold was canceled without a transfer timestamp
- Added a temporary dev-only override:
  - `DEV_INCIDENT_AUTO_CLOSE_MINUTES`
  - ignored in production
  - defaults to `30` when unset
- Documented the env var in `server/example.env`.

## Why

If an officer forgets to tap "I've left", the incident can remain open with no completion time even though all hold activity is finished. This change closes that gap while reusing the app’s existing timer/background-job patterns.

## Testing

Added coverage for:
- closing an arrived incident 30+ minutes after the last transfer
- not closing before the threshold elapses
- falling back to cancellation time when the last finalized hold was canceled without a transfer timestamp
- running the auto-close path through the existing scheduled `expireHolds` job

## Local Testing

Set this in `server/.env` for faster verification:

`DEV_INCIDENT_AUTO_CLOSE_MINUTES=1`

Then restart the local server container:

After transferring or canceling the last hold on an arrived incident, the incident should auto-close after about one minute and the UI should reflect the left/completed state automatically.

Closes #585


aaronhans (approved):
Nice Beaudry!

aaronhans:
This PR looks good to me but I wonder if we need more logging of things like this and where that should go. Added a note to https://github.com/sfbrigade/care-connect/issues/685 to consider

francisli:
@beaudrykock-moi @aaronhans @saadabdali I'm closing this PR because of the data model change, but leaving the associated issue open because I assume there's still a desire to automatically mark an officer as "I've left" after a timeout, but the implementation will be different now. I think it will probably be easier to start over than adapt this code.

Issue #605, opened 2026-04-10
Nudge user to confirm 849b narrative looks good before release


beaudrykock-moi:
So that the "edit narrative" option is clearer - testers were missing it.

AC:
In the confirm legal release view:
- [ ] Add a "I reviewed the narrative" checkbox below the edit narrative button; this does not persist to backend
- [ ] When "I reviewed the narrative" checkbox is marked, the Release reason section displays
- [ ] Confirm release button should not be enabled until the reviewed checkbox is marked and release reason accepted
- [ ] Change the "Edit narrative" button style to standard large blue CTA

Mocks is here — https://www.figma.com/design/Q8kS4FJXh1TbbM8eDR7Z6Y/CareConnect?node-id=7952-72860&t=9zv63PXyKxazR5V6-1

coilin:
<img width="932" height="863" alt="Image" src="https://github.com/user-attachments/assets/79a6334f-b950-484f-9d24-4d0568aef235" />

@beaudrykock-moi Proposing a slightly different solution here that keeps the intent of the story, but makes the flow feel a bit more explicit and less checkbox-driven.

Instead of adding an “I reviewed the narrative” checkbox, I’m suggesting we treat review as its own action inside the 849(b) card:

- Mark as reviewed — confirms the user has checked the narrative
- Edit narrative — remains available if they want to make changes
- after review is confirmed, the Release reason section is revealed
- Confirm release stays disabled until:
  - the narrative is marked as reviewed
  - a release reason is selected
  - the SFSO incident number is entered

After the review step is completed, the card switches to a completed state with:
- Reviewed
- Undo review

This still enforces the intended sequence before release, but feels more direct than a checkbox and gives the review step a clearer place in the flow.

Mocked it up here — https://www.figma.com/design/Q8kS4FJXh1TbbM8eDR7Z6Y/CareConnect?node-id=7952-72860&t=9zv63PXyKxazR5V6-1

Thoughts?

beaudrykock-moi:
Nice approach in general, I like it, thanks @coilin 

That said, I don't love "Mark as reviewed" (much like "Mark as complete" for safety checks). It's not descriptive of what the user is doing, which is confirming that the narrative looks good. But I'm struggling to find the right alternative language. "Looks good" is too casual, similar with "I reviewed the narrative". "Narrative reviewed" feels weird. 

So I guess "Mark as reviewed" it is. I think we can move to dev.

coilin:
[...]

PR #659, opened 2026-04-20
Adds sending to supervisor label to 849b review screen (#616)

### LINKED ISSUE #616: Make clearer that the 849(b) is sent to supervisor on release
For avoidance of confusion.

Current approach has it written in the page title, but it's the second sentence and physically separated from the action that does it, so many not be noted and/or understood.

Recommend we remove it from the title and instead aadd a note above/below the Confirm release button so that it's clear what happens.

Mocks — https://www.figma.com/design/Q8kS4FJXh1TbbM8eDR7Z6Y/CareConnect?node-id=7952-72860&t=9zv63PXyKxazR5V6-1

beaudrykock-moi:
## Summary
- update the SFSO legal release screen to match the Figma spec by moving the 849(b) supervisor warning into its own inline notice above the cancel/confirm buttons
- add the alert-circle icon
- simplify the header copy so the warning text appears only once

Closes #616

### DIFF
--- a/client/src/lesc/components/custody/LegalReleaseQuestions.jsx
+++ b/client/src/lesc/components/custody/LegalReleaseQuestions.jsx
@@ -130,7 +130,7 @@ function LegalReleaseQuestions () {
         <Stack gap='xl'>
           <Stack gap={0}>
             <Text size='xl' c='dimmed'>Confirm legal release</Text>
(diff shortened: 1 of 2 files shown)


beaudrykock-moi:
Going to fold this into #605. Closing this one.

PR #662, opened 2026-04-20
Adds 3 sec polling for facility and bed type data (#588)

### LINKED ISSUE #588: Chair capacity count and open/close status polling may be slow/not working
Actual:

- When care admin testers changed chair capacity or open/closed the facility, the chair count and facility status on other people's devices did not change unless the page was refreshed.

Expected:
- Changes to chair capacity or facility status should be communicated to other users quickly and without refresh

beaudrykock-moi:
## Summary

Fixes stale facility operational state across devices by polling facility status and bed-type data on the same 3-second interval already used for deflections.

## Problem

When a facility admin changed chair availability or updated the facility to closed/open-not-accepting, other users did not see the update until a manual refresh. The backend update succeeded, but the client kept using stale cached facility data because those queries only refetched on mount, focus, or reconnect.

## Root Cause

Deflection queries already poll every 3 seconds, but facility status and bed-type queries did not. Some screens also continued reading `facility.status` from the static facility context, which could stay stale for the full session.

## Changes

- Add shared live-query options for facility operational data with a 3-second refetch interval.
- Apply live polling to:
  - facility status banner queries
  - chair availability / bed-type queries in Care
  - chair availability / bed-type queries in Custody
  - facility status + bed-type queries in Manage Capacity
  - facility status + bed-type queries in Holds
- Update Holds to use the freshly fetched facility record when evaluating closed/open state and rendering facility details.

Closes #588

### DIFF
--- a/client/src/components/FacilityStatusBanner.jsx
+++ b/client/src/components/FacilityStatusBanner.jsx
@@ -5,6 +5,7 @@ import { IconAlertCircle, IconX } from '@tabler/icons-react';
 
 import Api from '@/Api';
 import { useFacilityContext } from '@/FacilityContext';
+import { facilityLiveQueryOptions } from '@/hooks/facilityLiveQueryOptions';
 
 function getBannerConfig (facilityName) {
   return {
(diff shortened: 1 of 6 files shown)


aaronhans (approved):
clean update

PR #668, opened 2026-04-21
Satisfaction survey

### LINKED ISSUE #82: As any user, I can report satisfaction with the experience
So that we can learn about needs and improve the app experience, as well as provide feedback to SFSO/DPH/SFPD leadership on RESET operations.

AC:

- [ ] All users get a satisfaction survey
- [ ] If any user rates 3 stars or less, prompt a follow up, Uber-style: "What could we improve about the app?" with an open text field - 200 char limit
- [ ] For both cases, support a "Can we ask a few more questions about RESET?" that links to one followup: "What could be improved about how the RESET center is working?" with an open text field - 200 char limit
- [ ] Should not ask for feedback every time. Once every 2 weeks or once every 10 holds, whichever is sooner.

traceylum1:
Closes #82 

CHANGES:
1. Created SatisfactionSurveyModal
2. Created useHook to handle different triggers (`SFPD` - leaving facility / `SFSO` - legal release / `CONNECTIONS` - medical intake)
3. Survey frequency of one month
4. Created endpoint to store survey responses anonymously, but store user dept (`SFSO`, `SFPD`, `CONNECTIONS`)
5. Added prisma model and migration
6. Added useHook mock, deflection tests, frontend modal tests


francisli (changes requested):
@traceylum1 this looks/works great with SFSO and CARE users...

However, with SFPD I'm getting 404 responses from the server when trying to submit feedback. I notice that the satisfaction-survey endpoint requires a deflection id to test for permissions? Once an officer has transferred their person, there's no longer a deflection in the list, so it's passing an empty string into the API call...?

Do we really need to have a deflection id authentication check, vs just checking to make sure a valid user is logged in?

francisli (approved):
Ok, lgtm now, tested all three org user types and was able to submit a survey response...!

PR #676, opened 2026-04-21
Shows incident details from hold add details not before (#582)

### LINKED ISSUE #582: Show incident details on first "add details" action, not first "add hold"
So the user isn't confused about being asked for details they may not be ready to provide.

AC:

- [ ] Incident details shows not when adding the first hold, but when first tapping "add details" on any hold in the incident
- [ ] Change label in incident view: "Start an incident" -> "Add incident details"
- [ ] Change button label: "Confirm incident details"
- [ ] Confirm button takes user to personal details entry, not back to home view

beaudrykock-moi:
## Summary

This PR updates the first-hold details flow for arresting users so incident details are collected later, when the LEO first chooses to add person details.

## Changes

- Allows the first hold to be placed immediately without showing the incident details form.
- Creates the incident in the background with blank nullable incident fields and an automatic arrest timestamp.
- When the officer first taps **Add Details** on an empty hold, routes them through the incident details confirmation form.
- Retitles that first incident-details form to **First, confirm incident details** to make it clearer this is a necessary first step.
- Changes the first-flow submit button to **Continue to person details**.
- After confirming incident details, sends the officer directly to the person details form for that hold.
- Skips initial red validation nudges on the first confirmation view, while preserving them for later incident edit views.
- Ensures canceling the cancel-incident modal does not change the confirmation-flow CTA (and renders icon-only controls without a `to` prop as buttons instead of router links, to prevent modal close button from accidentally affecting routing).

Closes #582

### DIFF
--- a/client/src/components/IconButtonLink.jsx
+++ b/client/src/components/IconButtonLink.jsx
@@ -4,6 +4,14 @@ import { Link } from 'react-router';
 import classes from './IconButtonLink.module.css';
 
 function IconButtonLink ({ to, variant, color, icon: Icon, onClick }) {
+  if (!to) {
+    return (
+      <Button className={variant === 'primary' ? '' : classes.default} color={color} classNames={classes} onClick={onClick} type='button'>
+        <Icon size={20} />
+      </Button>
+    );
+  }
+
   return (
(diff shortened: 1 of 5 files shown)


aaronhans (approved):
updated to fix merge conflict, looks good

PR #675, opened 2026-04-21
Increases default hold time to 90 minutes (#580)

beaudrykock-moi:
## Summary

- Increased the default hold duration from 60 minutes to 90 minutes.
- Updated manual hold extensions to reset active holds to 90 minutes.
- Updated reopened holds to use the same 90-minute expiration helper.
- Changed the database default for new deflection holds to expire after 90 minutes.
- Updated UI copy, tests, and fixtures to reflect the new 90-minute hold duration.
- Fixed the backend test helper to quote the Prisma schema path so tests can run from workspace paths containing spaces.

Closes #580

PR #678, opened 2026-04-21
Adds slight friction to hold button, avoid unintentional presses (#604)

### LINKED ISSUE #604: Add check if many holds added quickly
Feedback from user test: if an officer accidentally adds a bunch of holds, they may not realize it.

AC:

- [ ] Communicate a warning message to when if they hold 3 holds in quick succession
- [ ] Remind them they can cancel a hold if they created it accidentally and otherwise let them clear the modal and continue

Mocks: https://www.figma.com/design/Q8kS4FJXh1TbbM8eDR7Z6Y/CareConnect?node-id=14355-55179&t=6hqdaoS7HqHPl7Oz-1

beaudrykock-moi:
## Summary

Adds a small amount of friction to placing additional holds so users get clear feedback after tapping “Hold a chair” and are less likely to tap repeatedly.

## Details
- Shows a short `Placing hold...` state before creating an additional hold
- Disables repeat interaction while the hold placement delay/create request is in progress
- Uses the Figma-provided loader SVG and rotates it in CSS*
- Shows a success toast after hold creation: `Hold placed` / `1 chair reserved. Hold expires in 90 minutes.` (Note that this assumes #675 gets approved and merged.)
- Leaves the first-hold path untouched: if no active incident exists, the button still routes immediately to incident creation/details. This is to avoid clashing with the separate [PR](fix/582-incident-details-on-add-details) that changes the ordering on when the incident form first appears. May need some refactoring once that merges.

*Note on inline SVG
The closest Tabler/Mantine loader icon is different to the Figma loader glyph and in rotation does not visually match design. SVG inline is not great but matches design intent and avoids adding assets.

Closes #604

### DIFF
--- a/client/src/components/SunburstLoader.jsx
+++ b/client/src/components/SunburstLoader.jsx
@@ -1,0 +1,37 @@
+import { forwardRef } from 'react';
+import { Box } from '@mantine/core';
+
+import classes from './SunburstLoader.module.css';
+
+/**
+ * Radial/sunburst spinner from the Figma design, used for loading states on
+ * indigo-background buttons (e.g., "Placing hold..." in Holds.jsx).
+ *
+ * Fixed 20x20 and hardcoded white fill — intended for dark/primary surfaces.
+ * For other contexts, use Mantine's built-in <Loader> with an appropriate
+ * type/size/color.
+ */
+const SunburstLoader = forwardRef(function SunburstLoader (props, ref) {
+  return (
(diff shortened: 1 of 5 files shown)


aaronhans (changes requested):
We can use custom svg loaders in Mantine. If we do that we get less one off customization.

aaronhans (approved):
updated to add new loader component separately

PR #679, opened 2026-04-21
Adds scan button on legally released tab for SFSO users (#651)

### LINKED ISSUE #651: As an SFSO user, I can access the scan code button from the history tab view
So that users don't get lost and know they can access the scan button from anywhere

AC:

- [ ] Scan code button is visible regardless of the tab the user is on
- [ ] If scan succeeds, user is returned to active tab, not history 
- [ ] If scan fails or is canceled out, user is returned to history tab

beaudrykock-moi:
## Summary

- Keep the custody `Take custody` button visible on both `In custody` and `Legally released` tabs.
- Switch the custody view back to `In custody` after a successful transfer scan.
- Preserve the current tab when the scan flow is canceled or fails.
- Clean up: only fire transfer success for manual multi-code entry when at least one transfer actually succeeds.

Note: also made some changes to ensure tests were running locally; these may be duplicated in other PRs that I also worked on since I needed to clear tests on those also, so will require some clean up at merge.

Closes #651

### DIFF
--- a/client/src/lesc/components/custody/Custody.jsx
+++ b/client/src/lesc/components/custody/Custody.jsx
@@ -120,6 +120,7 @@ function Custody () {
   });
 
   function handleScanSuccess () {
+    setTab('in-custody');
     queryClient.invalidateQueries({ queryKey: ['deflections', facility.id] });
   }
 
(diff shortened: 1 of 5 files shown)


grahamplace (approved):
The button changes lgtm, the test changes I'm glossing over per your comment about cleaning up tests across PRs/at merge time

PR #686, opened 2026-04-21
Adds no chairs as cancel reason for holds (#681)

### LINKED ISSUE #681: As an arresting LEO, I can cite "No chairs available" as a reason when canceling a hold
So that if due to operational challenges, over-capacity has occurred but is not reflected in CareConnect's chair availability tracker, this can be tracked as impacting operations.

AC:

- [ ] Add "Over-capacity" to the list of available reasons an arresting LEO can choose from when canceling a hold before custody transfer

beaudrykock-moi:
## Summary

Adds `No chairs available` as a hold cancellation reason for arresting officers canceling a hold before custody transfer.

## Changes

- Added `no_chairs_available` / `No chairs available` to deflection cancel reason seed data.
- Added a Prisma migration to insert the new cancel reason into deployed databases.
- Added the new reason to DB test fixtures.
- Updated the cancel-reasons API test to assert the new reason is returned.

## Note for anyone running tests locally

If you're not seeing the changes and are confused (as I wasn't and was), remember that the cancel reason list is loaded from the database via `/api/deflections/cancel-reasons`. Existing localhost databases will not show the new option until the migration is applied or the database is re-seeded.

Closes #681

### DIFF
--- a/server/prisma/migrations/20260421000000_add_no_chairs_available_cancel_reason/migration.sql
+++ b/server/prisma/migrations/20260421000000_add_no_chairs_available_cancel_reason/migration.sql
@@ -1,0 +1,17 @@
+INSERT INTO "DeflectionCancelReason" (
+  "id",
+  "name",
+  "createdById",
+  "updatedById"
+)
+SELECT
+  'no_chairs_available',
+  'No chairs available',
+  "User"."id",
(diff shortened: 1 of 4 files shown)


aaronhans (approved):
looks good, clean data update

PR #699, opened 2026-04-23
Adds charge type selector to arrest details (#683)

### LINKED ISSUE #683: As an arresting LEO, I can select charge type
So that I can comply with legal reporting requirements.

AC:

- [ ] Somewhere in person details flow, arresting officer should be able to choose between "11550 HS" and "647(f) RWS" as the charge types. This is a required response.
- [ ] This must recorded on the 647(f) that is generated, in the "Charge" field (replace existing default).

beaudrykock-moi:
## Summary

Updates the arrest details step in the personal details forms to support selecting and persisting a charge type, align the page copy with the latest design, and use the selected charge on the generated 647(f) form.

## Detail

- Renamed the page from “Behavioral observations” to “Arrest details”
- Updated the subtitle to: “Describe what you observed that justifies the arrest, and select charge type.”
- Renamed the textarea label from “Arrestable behavior” to “Behavioral observation”
- Added a required “Select a charge type” question with chip options:
  - `11550 HS`
  - `647(f) RWS`
- Added `chargeType` to completion gating so personal details remain incomplete until it is answered
- Persisted the selected charge type on `Deflection`
- Updated hold/details surfaces to display the saved charge type
- Updated 647(f) generation to use the saved charge type instead of the previous hardcoded default
- Matched the arrest-details screen more closely to design tokens:
  - title uses Mantine `h3` equivalent
  - subtitle uses Mantine 20px token equivalent
  - helper text spacing under the textarea tightened to 4px token spacing
  - spacing between the charge section and next button set to `spacing-xxl`
- Changed the audio recorder CTA text from “Record” to “Record voice” (unrelated to this ticket but needs fixing anyway)

Closes #683


aaronhans (approved):
Nice update!

Tiny commit added to fix accessibility issue

Could use enum for charge type but fine with only two

francisli (approved):
I agree with @aaronhans that the column should be an enum, I made the change and will now merge in...

PR #702, opened 2026-04-23
Handoff security

### LINKED ISSUE #600: custody chain integrity edge case: handoff
There are some unlikely scenarios that could result in confusing situations. If we make the extra effort to guard against these scenarios we avoid causing problems for officers in the field.

## Block QR Code Replay

Code problem: QR codes contain just the deflection ID (/handoff/42). They can be screenshotted, photographed, or shared. Each scan successfully transfers custody to the scanner.

Potential fix: After a successful handoff, reject subsequent scans. Require generating a new QR code for further handoffs. This is a server-side change only — add a check that currentOfficerId matches the expected "from" officer before allowing the handoff.

## Block duplicate simultaneous scans

Code problem: The handoff endpoint fetches the deflection OUTSIDE the transaction (line 33), performs validation checks, then updates inside the transaction WITHOUT re-fetching. Every other status transition endpoint (transfer, admit, release, safety-check) acquires a row-level lock via bedType.findByIdForUpdate() and re-fetches the deflection inside the transaction. Handoff does neither.

Potential fix: Add lock acquisition and deflection re-fetch inside the transaction, matching the pattern in transfer.js. Add one-time-use enforcement so the second scan is rejected.

aaronhans:
Defense against unlikely edge cases:

- Re-scanning an existing QR handoff code
- Concurrent scan of same QR handoff code


francisli (approved):
Now that we're introducing row-level locking on Deflection, I wonder if we need to review whether there are similar reentrant/concurrent edge cases that we should acquire locks. Initially, I only implemented locks on BedType when updating a deflection modifies its cached counts. However, it might make sense to always also acquire a lock on the deflection itself that is being modified... needs some more thought...

PR #713, opened 2026-04-24
SFPD: Improve the History view

### LINKED ISSUE #694: Bug: SFPD user can see QR code for handed-off person in History view
- Officer Alice detains someone
- Alice hands the person off to Officer Bob
- Bob arrives at RESET. In his Active Holds, he now sees the Custody Transfer QR code.
- **Actual behavior:** Alice looks at her History view, and _also_ sees the QR code there
- **Expected behavior:** Alice should see the hold in her History, labeled as `Handed off`, but she shouldn't see the QR

More generally: I haven't looked at exactly how we do this now, but our card view for a hold should be sensitive to the context in which it's being displayed (we could pass that in as a prop). Active holds should be displayed a certain way, and History holds will have other behaviors.

### LINKED ISSUE #704: SFPD: Productionize History view
The History view works at low volume, but looks messy to me.

## Current behavior
- Three parallel queries to `GET /api/deflections` (`inactive`, `post-transfer-active`, `handed-off`), each refetched every 3s.
- then the client also appears to call `incidents/{id}` for every single incident associated with those holds
- Server paginates the holds (`perPage=25` default) but the client never requests page 2+ and ignores pagination headers.
- Once a bucket exceeds 25 rows, older items silently disappear — no "load more", no error.
- Client-side merging of the three buckets makes chronological ordering fragile.

## Proposed changes
- Collapse into a single `list` call which returns all the History info the client needs (with incident info bundled as needed)
- Consider dropping history polling. (Could fetch history updates only when needed, when data in `my-holds` changes?)

### LINKED ISSUE #714: SFPD: In History view, let me click View Details on a Hold I handed off
In the History view, typically I can click 'View details' on a hold card to see a (read-only) view of the incident and subject details.

For some reason we don't show the button on Handed Off cards. Probably this is just an oversight.

Image

Currently, if I use the URL bar to navigate directly to the relevant Hold:
- the UI looks like I can edit the Hold
- the backend prevents me from saving edits
- If we were offering this link directly, we'd want to show the read-only view I think.

saadabdali:
NOTE: I built this PR on top of my large branch `sa/655-incident-hold-refactor`. We should switch it to target `dev` before merging. 

Closes #704
Closes #694
Closes #714 

## In this PR
- We simplify the way we fetch data for the History tab. Previously, we fetched history using three different calls and combining their results. Now, we make a single call, reduce the frequency of calls (by killing unnecessary polling), and can use the results directly.
- Enable an SFPD officer to `View Details` from the History page for a Hold that was handed off. (And ensure that he sees a read-only view of the resulting page; he shouldn't be editing the hold if he's not the current officer.)
- Stop calling the cancel-reason endpoints to render History cards (don't need to)
- Fix a bug where a Field user who handed off a Hold can still see the Transfer QR in the History view

---
# Test Plan 
### A.1 Landing on the Holds screen (Active tab)
- [x] `/api/facilities/:id/my-holds` polls every ~3 s.
- [x] **No** `/api/deflections` requests polling at 3 s. Exactly one fires
      at mount (for history).
- [x] **No** `/api/incidents/:id` follow-up requests after that history
      fetch returns.

### A.2 Switching to History tab
- [x] No new network activity. History was fetched at mount, tab switch
      is a no-op.

### A.3 Flipping between tabs repeatedly
- [x] No new history fetches. `my-holds` keeps ticking; nothing else.

---

### B.1 Inspect the history request
- [x] One call to `/api/deflections?scope=history&includeIncident=true&perPage=100&facilityId=...`
      (plus `page=1` if present).
- [x] Response body: each row has `incident` embedded with address,
      `arrestedAt`, etc.
- [x] Response contains (in one array):
  - inactive holds with a subject attached,
  - post-transfer-active holds you own or created,
  - active holds you created but handed off to someone else.

---

### C.1 Custody takes a hold
- [x] Hold a chair, fill everything, arrive, have Custody scan the QR.
- [x] Within ~3 s, the hold disappears from Active.
- [x] A new `/api/deflections?scope=history...` call fires.
- [x] History tab shows the hold with the "Transferred" badge.

### C.2 Officer cancels a hold
- [x] On Active tab, cancel a hold.
- [x] Within ~3 s, a new `/api/deflections?scope=history...` call fires.
- [x] History shows the cancelled row with "Canceled at …".

### C.3 Hold expires
- [x] In psql, set a hold's `expiresAt` to the past
- [x] Within ~3 s, hold disappears from Active.
- [x] History refetch fires.
- [x] "Expired" badge on the row.

### C.4 Handoff out
- [x] Initiate a handoff on a hold; have another officer scan the code.
- [x] Within ~3 s, the hold disappears from your Active tab.
- [x] History refetch fires.
- [x] Row appears in History with "Handed off" badge.

---

### D.1 Idle
- [x] Sit on the History tab for 30 s with nothing changing.
- [x] **Zero** `/api/deflections` calls in that window. Only `my-holds`
      polls.

### D.2 Idle on Active tab
- [x] Sit on the Active tab for 30 s with nothing changing.
- [x] Same: zero `/api/deflections` calls after the initial mount fetch.

---

### E.1 Grouping
- [x] Holds from different incidents appear in separate `IncidentGroup`
      cards, each header showing address + `arrestedAt` pulled from the
      embedded incident.

### E.2 Badges
- [x] Transferred (teal "Transferred")
- [x] Handed off (indigo "Handed off")
- [x] Cancelled ("Canceled at HH:MM" with reason in parens if applicable)
- [x] Expired ("Canceled after expiry")

### E.3 Empty state
- [x] Fresh account with no history shows "You don't have any past
      holds."

### E.4 Click-through
- [x] Clicking a history row navigates to `/holds/{id}`.

---

### F.1 Sender's perspective
- [x] Alice creates → fills → hands off to Bob. Bob transfers to custody.
- [x] Alice's History shows the hold with "Handed off" badge (not
      "Transferred" — that would be Bob's perspective).

### F.2 Receiver's perspective
- [x] Same setup as F.1.
- [x] Bob's History shows the hold with "Transferred" badge.

### F.3 No QR on sender's history row (issue #694)
- [x] Alice hands off to Bob. Bob arrives at RESET but hasn't transferred
      yet (hold still `ACTIVE`, subjectStatus `ONSITE_AWAITING_TRANSFER`).
- [x] On Alice's History tab: row shows "Handed off" badge.
- [x] **No QR code rendered on Alice's history row.**
- [x] No "Transfer code: NNN" text on Alice's history row.
- [x] Bob's Active tab still shows the QR code (that's his to scan).

### F.4 /holds/{id} is read-only for non-owner
- [x] Alice hands off a hold to Bob. From Alice's History tab, tap
      "View Details" on that row → lands on `/holds/{id}`.
- [x] **Verify:** page renders hold info, but NO Edit buttons anywhere
      (Edit details, Edit narcotics, Edit substance details, Edit
      arrest, Edit property, Edit incident).
- [x] **Verify:** no bottom action footer (no "Cancel hold" or
      "Finish details" button).
- [x] From Bob's Active tab, tap the same hold → lands on `/holds/{id}`.
- [x] **Verify:** Edit buttons are present. Bottom action footer is
      present (since Bob owns the hold).
---

### G.1 Pre-subject cancellations excluded
- [x] A hold that was cancelled before any subject details were entered
      does **not** appear in History.


francisli (approved):
Thanks for this refactor- we really need to review the implementation across ALL roles and tabs, not just SFPD History, to reduce the number of calls, particularly to the deflections list endpoint.

Originally I was hoping to just use the "active" boolean parameter representing the two tabs of the main view and then the roles to define the scope of the queries (i.e. active = true is "Active holds", active = false is "History" for SFPD, etc)...

PR #732, opened 2026-04-24
Run server tests concurrently via shared containers

grahamplace:
# Summary

Speeds up the server test suite by starting Postgres + MinIO once up front and running `node --test` with parallelism enabled, instead of spinning up a fresh container pair per test file serially.

## Changes

- New `server/test/run-tests.js` orchestrator: starts shared Postgres + MinIO containers once, writes connection info to a tmpfile, discovers `**/*.test.js`, splits them by whether they call `mock.module(...)`, and runs them in two passes against the same containers before tearing down on exit.
- Two-pass execution avoids a Node loader race: `--experimental-test-module-mocks` patches the CJS resolver and was racing against Prisma's `#main-entry-point` subpath import under parallel loads, surfacing as flaky `Cannot find module '#main-entry-point'` failures.
  - **Pass A** — files that don't call `mock.module()` (~46 of 50 today). Runs at the configured concurrency (default 8) with the standard, unpatched loader. No experimental flag, no race.
  - **Pass B** — files that call `mock.module()` (4 today: `posthog-error.test.js`, `posthog.test.js`, `generateForms.test.js`, `forms.test.js`). Runs with `--experimental-test-module-mocks` at concurrency 1, so the patched loader is on but nothing is racing against it.
  - Both passes run unconditionally; final exit code is `passA || passB`. Routing is auto-detected by regex (`\bmock\.module\s*\(`) so new mock-using tests are picked up without maintaining an allowlist.
- `server/test/helper.js`: reads shared container info from `CARE_CONNECT_TEST_CONTAINERS`; generates a unique per-file DB + template DB; `recreateDb()` uses `CREATE DATABASE ... TEMPLATE ...` for fast reset. Falls back to per-file containers when invoked outside `run-tests.js`.
- `server/package.json`:
  - `npm test` → `node test/run-tests.js` (the orchestrator owns the flag and concurrency decisions per pass; concurrency overridable via `TEST_CONCURRENCY`).
  - `npm run test:sequential` preserves the previous fully-serial behavior as an escape hatch.
- Shared-container startup in `run-tests.js` honors `TESTCONTAINERS_NETWORK_MODE` so CI parity is preserved.
- Switches all `@prisma/client` imports to a `prismaPkg` default-import + destructure pattern so production code itself does not depend on the experimental loader for ESM/CJS interop.

## Concurrency-induced flake fix in `left.test.js`

The "remains consistent when left races with cancellation of the last active arrived hold" subtest in `server/test/routes/api/facilities/left.test.js` (added in #737) asserted that both responses in its `Promise.all` race must return 200. Under heavy parallel test load, `/left` consistently commits before cancel's pre-lock auth read in `cancel.js`, which makes cancel correctly return 403 — the caller is no longer the hold's `currentOfficer`, so `canModifyDeflection` rejects. On dev with `--test-concurrency 1` the system is idle enough that cancel's read wins and both return 200; under our concurrency=8 we cross a load threshold and the timing flips.

Both race outcomes are valid: the post-race invariants the test then checks (no active arrived holds, `myHolds.atFacility=false`, DEPARTURE check-in present) hold either way, and that's what the test name actually claims to verify. Relaxed the cancel assertion to accept 200 *or* 403; production code is unchanged.

# Explanation

Copy pasting this explanation which is a good summary of the change in approach here:

## Before

Each test file was self-contained. When `helper.js`'s `build(t)` ran — every test file calls this in `beforeEach` / setup — it did the full setup from scratch:

1. Spin up a brand-new Postgres testcontainer.
   - Docker pulls the image, starts a container, and waits for readiness.
   - This adds roughly 5–10 seconds of overhead per file.

2. Push the Prisma schema into it.
   - `prisma db push` runs all the schema DDL into the new database.
   - This takes several seconds.

3. Load all DB fixtures from `server/test/fixtures/db/*.yml`.
   - Every user, facility, deflection, and related record that tests rely on gets loaded.
   - This takes a few seconds.

4. Spin up a brand-new MinIO container.
   - Same process: fresh container, wait for readiness.

5. Build the Fastify app against that fresh DB and storage backend, then hand it to the test.

6. At test end, tear down both containers.

Tests ran serially with:

```bash
--test-concurrency 1
```

With roughly 50 test files, each paying that startup tax sequentially, the suite was dominated by container startup rather than actual test logic.

This setup was correct — perfectly isolated per file — but extremely slow.

## After

The orchestrator, `server/test/run-tests.js`, flips the model:

- shared containers
- parallel files
- fast per-file DB resets

The new flow:

1. Start one Postgres container and one MinIO container once before any test runs.
   - Their connection info is written to:

   ```txt
   /tmp/care-connect-test-containers.json
   ```

2. `helper.js` reads that file instead of starting its own containers.
   - If the env var is not set, such as when someone runs a single test file with `node --test`, it falls back to the old per-file-container behavior.
   - This preserves single-file dev workflows.

3. Each test file gets a unique database inside the shared Postgres instance, plus a template copy of that database.
   - The schema is pushed once per file into the template.
   - Fixtures are loaded once per file into the template.

4. Resets between tests use Postgres-native database cloning:

   ```sql
   CREATE DATABASE ... TEMPLATE <template_name>
   ```

   This is a file-system-level copy and is much faster than re-running `prisma db push` and re-inserting fixtures.

5. Test files run in parallel.
   - With `--test-concurrency 8`, eight test files execute concurrently as separate Node processes.
   - Each file has its own database inside the shared Postgres container, so they do not step on each other.

6. A single MinIO bucket is shared across files.
   - Storage keys are namespaced per test, so storage interactions remain isolated.

7. A two-pass split was added today.
   - The orchestrator regex-scans test files for `mock.module(`.
   - The 4 files that need it run in a serial Pass B with:

   ```bash
   --experimental-test-module-mocks
   ```

   - The other roughly 46 files run in a parallel Pass A without the flag.
   - This avoids the Node loader race that was causing flaky `Cannot find module '#main-entry-point'` failures.

## What That Buys You

- Container startup tax is paid once, not 50 times.

- DB reset is now:

  ```sql
  CREATE DATABASE ... TEMPLATE
  ```

  instead of:

  ```txt
  drop, recreate, push schema, reload fixtures
  ```

  This makes per-test resets orders of magnitude faster.

- Parallelism compounds the win.
  - With 8 files in flight, the suite walks through the file list roughly 8× faster than before, even on a 4-vCPU runner.

- Reliability improves.
  - The two-pass split means the experimental loader flag only touches the small set of files that genuinely need it.

## Net Effect

The suite used to be dominated by infrastructure churn.

Now it is dominated by actual test work — and reliably so.


francisli (approved):
THANK YOU for working on this! Not sure I understand yet why the Prisma package imports need to change, but I'll dig into it later...

Issue #721, opened 2026-04-24
Incident left can race with concurrent hold state transitions


beaudrykock-moi:
Problem

`POST /api/facilities/:facilityId/left` can race with concurrent hold state transitions and leave officer/facility-presence state inconsistent with the final set of active holds.

Current route behavior

`POST /api/facilities/:facilityId/left`:
- clears `currentOfficerId` on active holds at the facility where the officer has already arrived
- records a `FacilityCheckIn` departure event

Potential race

If `left` runs concurrently with hold cancellation, transfer, release, reopen, or other hold state changes:
- the set of holds matched by `left` can change during execution
- departure can be recorded against a stale snapshot of “holds at facility”
- final officer presence and hold ownership state may not match what actually remained active after both operations settle

Impact

- officer presence can become inconsistent with the final hold set
- `my-holds` / at-facility workflows can show stale or contradictory state
- check-in/check-out reporting may not reflect the actual final transition set

Acceptance Criteria

- `POST /api/facilities/:facilityId/left` must be resilient to concurrent hold-state transitions
- departure must only clear `currentOfficerId` for holds that are still eligible at commit time
- if no eligible holds remain by commit time, the route must still behave consistently and not leave partial/stale officer-presence state
- facility departure records must remain consistent with the final hold set after concurrent transitions settle
- add regression coverage for concurrent `left` vs hold cancellation / transfer of the last active hold

Issue #720, opened 2026-04-24
Unsafe Transaction Control Flow via reply.send() Inside Prisma Transaction Callbacks


beaudrykock-moi:
### Problem
Several routes call reply.code(...).send() from inside fastify.prisma.$transaction(async (tx) => { ... }) callbacks instead of throwing an error and handling the response after the transaction exits. This creates a risky control-flow pattern where the HTTP response can indicate failure while the transaction still commits earlier writes, or where partial state changes can slip through under concurrency.

### Why this matters
A Prisma transaction callback should either:

- complete successfully and commit, or
- throw and roll back

Returning an HTTP response object from inside the callback does not reliably express “abort this transaction.” It only exits the callback path. If earlier DB writes already happened, they may still commit.

### Evidence in current code

- Confirmed bug in incidents/create.js
   - losing requests return 410, but Incident and IncidentOfficer rows still persist

- Same pattern also appears in routes such as:
   - deflections/create.js
   - deflections/reopen.js
   - facilities/_facilityId/bed-types/update.js

### Impact

- Partial commits on logical failure
- Response/DB mismatch
- Concurrency bugs become easier to trigger and harder to reason about
- Rollback behavior depends on fragile control flow instead of explicit transaction semantics

### Acceptance Criteria

- No route may call reply.send(), reply.code(...).send(), or otherwise write the HTTP response from inside a Prisma transaction callback.
- Transaction callbacks must use one of these patterns only:
   - throw a typed/domain error to force rollback
   - return plain data to the outer scope, with response construction done after the transaction finishes
- Logical failure branches inside transactions must throw, not return reply objects.
- Existing affected routes are refactored to the safe pattern.
- Add regression coverage proving that failed transactional requests do not persist partial writes.
- Add a code search or lint/test safeguard to prevent reintroducing reply.* inside Prisma transaction callbacks.

Issue #719, opened 2026-04-24
Incident cancel can complete while new hold is concurrently added


beaudrykock-moi:
Problem

`POST /api/deflections` can attach a new active hold to an incident even when that incident should be treated as closed to further hold activity.

Specific failure mode

- deflection creation validates facility state and bed availability
- it does not validate any incident-level eligibility inside the transaction
- if the product treats some incidents as no longer eligible for new holds, that rule is currently unenforced

Impact

- new holds can be attached to incidents that operators consider finished
- UI and permissions can diverge from persisted hold state
- incident-level workflow assumptions are not reliable

Acceptance Criteria

- define an explicit incident-level eligibility rule for new holds
- enforce that rule atomically inside deflection creation
- under concurrency, deflection creation must not attach a new hold once incident eligibility has transitioned to disallow new holds
- add regression coverage for the relevant race

beaudrykock-moi:
I think this is moot given Saad's recent changes, but will flag for review.

Issue #718, opened 2026-04-24
Bed type update and facility close can double-decrement the same hold


beaudrykock-moi:
### Problem
Concurrent bed-type shrink and facility close can both process the same in-transit hold, causing negative holds and inTransit counters.

### Evidence
Load test: bed-type-update-vs-facility-close
Result:
expected holds: 0, inTransit: 0
actual holds: -1, inTransit: -1
Routes:
[bed-types/update.js](app://-/index.html#)
[facilities/status.js](app://-/index.html#)

### Specific failure mode
- Both routes independently find active DETAINED holds and cancel them.
- Neither route re-fetches/validates each hold after locking the bed type to ensure it is still active before decrementing counters.
- Both routes subtract counts based on stale assumptions.
- Facility close also decrements without [Math.max](app://-/index.html#), at lines 112-117.

### Impact
- Impossible negative bed counters
- Broken bed availability math
- High-risk integrity corruption in admin workflows

### Acceptance Criteria

- Bed counters must never become negative under any concurrent admin operation.
- A hold may only be cancelled and decremented once, even if multiple admin operations race.
- Both facility close and bed-type shrink must re-check hold status after lock acquisition before applying decrements.
- Under concurrent bed-type-update vs facility-close:
   - holds >= 0
   - inTransit >= 0
   - bed counts reconcile to actual active deflections
- Add regression coverage for this race.

PR #737, opened 2026-04-25
Fix/729 multiple race conditions

beaudrykock-moi:
## Summary

Fix the Prisma deflection test seed and harden incident/deflection/facility workflows against stale reads and race conditions.

This bundle updates the test seed to match the current schema, then consolidates a broader pass of transactional correctness across incident creation, hold creation/reopen/cancellation, facility arrival/left flows, admit, bed-type updates, subject upserts, property return, handoff, and extension paths. The main change is to move validation and state transitions inside Prisma transactions under the relevant row locks, ensure transaction callbacks only commit or throw, and prevent losing concurrent requests from persisting partial or duplicate state.

## Changes

- fix `server/prisma/seeds/testDeflections.js` to stop writing removed lifecycle fields on `Incident` and write `arrivedAt` on `Deflection`
- `#715` make incident bed allocation atomic and roll back fully on contention/failure
- `#716` block hold creation when the facility is not `OPEN_ACCEPTING`, with checks performed under lock inside the transaction, and prevent facility close/create races from leaving an active in-transit hold on a closed facility
- `#717` block deflection reopen when the facility is not `OPEN_ACCEPTING`, with validation performed under the facility lock inside the transaction
- `#718` prevent close/shrink races from double-cancelling holds or driving facility counters negative
- serialize admit transitions on the deflection row so concurrent admit requests allow only one `READY_FOR_INTAKE -> ADMITTED` transition, write one audit row, and return `409` to the loser
- `#720` standardize transactional handlers so HTTP responses are emitted only after the transaction exits, making rollback semantics explicit and safe
- `#721` make facility `arrived` and `left` operate on locked commit-time hold state so cancellation/transfer races do not leave stale officer-presence or facility-presence state
- serialize deflection subject, property-return, handoff, and extend updates so concurrent requests cannot create duplicates or apply stale writes
- re-check hold state at cancel/update time so counters, audits, and side effects are derived from commit-time state only

## Why

The seed had drifted from the schema, and several routes were still making decisions from pre-transaction snapshots or responding from inside transaction callbacks. Under concurrency, that allowed partial writes, stale updates, duplicate records, and inconsistent counters. This change makes rollback behavior explicit and aligns all affected flows with commit-time state.

## Testing

- `node --experimental-test-module-mocks --test-concurrency 1 --test test/routes/api/incidents.test.js test/routes/api/deflections.test.js test/routes/api/deflections-handoff.test.js test/routes/api/deflections-extend.test.js test/routes/api/facilities/arrived.test.js test/routes/api/facilities/left.test.js test/routes/api/facilities/my-holds.test.js test/routes/api/facilities/hold-cancellation.test.js test/routes/api/facilities/bed-types.test.js test/no-reply-in-transaction.test.js`

`#719` was moot with the changes to incident structure; no fixes were made.

Closes 715, 716, 717, 718, 719, 720, 721, 735.


aaronhans:
reviewing test failures

saadabdali:
Aaron and I were both tagged as reviewers - I don't want to step on any toes, but if it's helpful, am happy to look at this today and get it merged in some form by end of day.

beaudrykock-moi:
@saadabdali I tagged you both b/c I felt especially uncomfortable with these changes and felt two sets of expert eyes might be better than one! Sorry to confuse things. If you have time today to review and get things merged ASAP, I think you should go for it - but @aaronhans feel free to coordinate with Saad. Thanks both!

aaronhans:
Thanks @saadabdali! These changes are really complex so would benefit from a lot of review. I added a little update to it to cover an uncovered case. I saw that this brand was passing an updated version of all the tests in https://github.com/sfbrigade/care-connect/pull/722 but now we have some issues with latest dev merge

saadabdali:
# Review notes

I made a bunch of changes to this PR, and want to add my review notes and comments separately.

## Concurrency protections by route
- `incidents/create.js` (#715) — obtain bedType lock before creating incident; #720 remove `return reply` from tx
- `deflections/create.js`, `reopen.js` (#716/#717) — obtain facility lock before validating facility status
- `bed-types/update.js`, `facilities/status.js` (#718) — serialize shrink/close operations under facility lock; prevent double-cancellation and negative counters
- `admit.js` (#735) — obtain deflection-row lock, preventing double-admit
- `subject.js` — obtain deflection lock before checking subjectId. prevents orphaned Subject rows.
- `property-return.js` — use deflection lock to prevent duplicate property-return audit rows under concurrent submissions
- `arrived.js`, `left.js` (#721) — re-check eligibility at commit; operate under facility lock


## Potential future work

- **#781** — This PR originally included a brittle test for #720 that I thought we shouldn't merge; I removed it  and punted that test to a separate ticket
- `reopen.js` declares its own `noAvailableBedError()` instead of using the shared one. Left a `TODO` indicating we should unify this in the future. (I didn't want to change the return format in this PR; it was big enough already.)
- **Wire-format inconsistency** — routes with strict schemas (transition handlers, handoff, property-return) keep their shapes; routes without unified to Fastify default. Punted on full unification for now.

## Non-obvious decisions

- **`admit.js` locks the deflection row, not the bedType.** Most transition handlers lock the bedType. This one doesn't because  `admit` doesn't change bed counter (so we just lock the deflection itself). 
- **Removed pre-lock validation reads**. Across several files, I opted to remove the double-read pattern (where you first read the row for initial validation, then obtain a lock, then re-read the row). I did this for clarity - future readers of the code might be confused about why we're performing the validation twice, or modify the validation logic in one place but not both (this had already happened in the initial PR). This is debatable.
- **`reopen.js` dropped its `catch + translate`** wrapper; its 4xx responses now use Fastify's default body (4 reopen tests updated).

saadabdali (approved):
I made a bunch of changes to this PR, so will pass to @francisli for a once-over rather than merging directly.

francisli (approved):
Going to merge this in now, although I think that per the conventions and patterns that we're establishing, we should consider locking deflection records for all updates- including those where we first lock bedType.

PR #743, opened 2026-04-26
Added second SFSO user to seed

beaudrykock-moi:
## Summary
Add a second SFSO test user to the Prisma seed data.

## Changes
- Add `sfso2@careconnectsf.org` to `server/prisma/seeds/users.js`
- Mirror the existing SFSO test user roles and organization settings
- Assign a unique star number for the new seeded account


server/prisma/seeds/users.js
```diff
@@ -30,10 +30,20 @@ export default async function main (prisma) {
...
       badgeNumber: '5678',
       mfaEnabled: false,
     },
+    {
+      email: 'sfso2@careconnectsf.org',
+      firstName: 'SFSO2',
+      lastName: 'User',
+      isAdmin: false,
+      organizationId: 'sfso',
+      roles: ['FIELD', 'CUSTODY', 'ORG_ADMIN'],
```
saadabdali: The PR description seems to say we want to mirror the settings of the existing SFSO user, but this line gives the new SFSO2 user `FIELD` access in addition to `CUSTODY`.

Is that intentional? I know we intend to support that flow very soon (allowing users with both roles to choose which role to work in). I wasn't sure if we had already built that functionality.

grahamplace: Yes, we merged it in today and the new intention is that ALL SFSO users have `FIELD` and `CUSTODY` by default. [Discussion](https://sfbrigade.slack.com/archives/C09PJA0E77G/p1777219050414789?thread_ts=1777066587.369309&cid=C09PJA0E77G)

I do think we might want to remove `ORG_ADMIN` from this new user, just so we have a variety of seeded users to test with (the current SFSO user already has `ORG_ADMIN` role)

grahamplace (approved):
lgtm % my comment about possibly removing `ORG_ADMIN` role from this user. Not a blocker though

beaudrykock-moi:
Yeah sorry @grahamplace I first did this off your PR b/c wanted to test with the code, so it did mirror the structure there, but then I moved it to a new branch and forgot to adjust.

I'll remove the org-admin. Good call.

saadabdali (approved):

PR #747, opened 2026-04-27
Updates release reasons per SFSO request (#731)

### LINKED ISSUE #731: Add/adjust release reasons
For SFSO, edit the available release reasons as follows

- Medical issue -> Medical issue (physical)

Add:

- Behavioral health evaluation

No other changes to flows. Just one edit and one addition.

beaudrykock-moi:
## Summary

Updates the custody legal release reason options to support the new release workflow.

## Changes

- Renames `Medical issue` to `Medical issue (physical)` in the custody legal release UI
- Adds `Behavioral health evaluation` as a new selectable release reason chip
- Updates release reason seed data and test fixtures so the new option persists via `releaseReasonId` - these are very targeted; decided not to make them cover just the rendering of the changes and not all the options, but can easily change that if needed
- Extends client and server tests to cover the renamed label and new behavioral health release path

Closes #731


francisli (approved):
Approved, although I'm thinking we should consider dropping some of these table-backed enum record types and just replacing them with db enums. They're not being dynamically loaded anywhere, but manually placed into specific places in the UI so we're just adding some extra overhead with their management.

PR #757, opened 2026-04-27
[Dev] Update test users so they look distinct

### LINKED ISSUE #752: It's hard to tell our test users apart
In the app header, all of our users now show up as `S. User`. 

I used to see: `SFPD User`, `SFPD2 User`, `SFSO User`, etc.
Now *all* of these users show up as `S. User` which is needlessly confusing if you're a developer playing multiple roles in multiple browsers.

I assume that abbreviating the first name is a deliberate recent UI change. In that case, I'll just rename the test users so that they'll remain visibly distinct in the app header.

saadabdali:
Closes #752.

Before:
Image

After:
Image

In this PR:

- Update the seed files to change the display names for the test account. This way, the test accounts (used by developers) will actually look distinct from each other. Currently they all look exactly the same (`S. User`). This happened after we changed the app header to only display the first initial rather than the full first name. (We apparently did this in order to make room for the role badge that we sometimes need to show.)
- The usernames and passwords are all unchanged. Only the display `firstName` and `lastName` are changed here.

Deployment notes:

- Nothing to do in prod, since (I assume) we do not run this seed on Prod.
- Developers will need to re-seed their local DB in order to see the change.

### DIFF
--- a/server/prisma/seeds/users.js
+++ b/server/prisma/seeds/users.js
@@ -6,8 +6,8 @@ export default async function main (prisma) {
   const users = [
     {
       email: 'sfpd@careconnectsf.org',
-      firstName: 'SFPD',
-      lastName: 'User',
+      firstName: 'Test',
+      lastName: 'SFPD1',
       isAdmin: false,
       organizationId: 'sfpd',
       roles: ['FIELD'],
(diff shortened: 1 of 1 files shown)


grahamplace (approved):
Thanks! I should've done this on my PR. Yep, I picked up using first initial from Design's changes for the "work mode" stuff which included the badge and thus needed a shorter name in the header

PR #759, opened 2026-04-27
Allow valid emails with apostrophe

### LINKED ISSUE #754: Allow apostrophes in email address validation
This problem was seen in real life at today's RESET simulation.

Apparently a valid email address can contain an apostrophe (in the local part of the address, i.e. left of the `@`). An officer was something like `po'rourke@something.gov` and the app was refusing to accept the email address.

https://www.reddit.com/r/techsupport/comments/1nhn6r/can_an_email_have_an_apostrophe_in_it/

saadabdali:
Closes #754 

The client and server both use regexes to validate email addresses. 
Client version is built-in to Mantine. 
Server version is built-in to Zod.

The two regexes disagree. Zod allows apostrophe in the name portion of an email address, which is [technically correct](https://www.reddit.com/r/techsupport/comments/1nhn6r/comment/cciojxw/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button). Mantine does not.

In this PR:
- Create a new shared util for email validation
- Have all clients use that util function instead of Mantine's built-in validation
- Add a test which covers this case (and a bunch of other common accept/reject cases for email validation)

The relevant context I was solving for was the `Invites` page (`/admin/invites/new`) in the Admin UI, but this change coversa bunch of other client forms (registration, login, etc)

Note: I originally intended to _exactly_ copy the regex used by Zod into the client-side email util. Unfortunately our linter doesn't like it (it has a few unnecessary escape chars). The hope is that while they are no longer literally identical, the client and server email validation still  have the same behaviors.

### DIFF
--- a/client/src/Admin/Invites/AdminInviteForm.jsx
+++ b/client/src/Admin/Invites/AdminInviteForm.jsx
@@ -1,6 +1,8 @@
 import { useNavigate } from 'react-router';
 import { Alert, Button, Checkbox, Container, Fieldset, Group, Stack, Select, Textarea, TextInput, Title } from '@mantine/core';
-import { isEmail, isNotEmpty, useForm } from '@mantine/form';
+import { isNotEmpty, useForm } from '@mantine/form';
+
+import { isEmail } from '@/utils/email';
 import { useMutation, useQuery } from '@tanstack/react-query';
 import { Head } from '@unhead/react';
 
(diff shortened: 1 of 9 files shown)


saadabdali:
Switching back to draft because there's a lint error - sorry!

francisli (approved):

PR #761, opened 2026-04-27
PWA with minimal-ui and smallest serviceworker

aaronhans:
Create Progressive Web App 

- uses icons from design team (official orange and blue logo was considered and not chosen for this purpose)
- Chooses minimal-ui option which will use most like normal browser chrome but how this is interpreted varies across OS
- Uses smallest possible service worker


francisli (approved):
This Vite PWA plugin looks cool, let's get this in and try it out..!

PR #802, opened 2026-04-29
Removes exit question and adjusts form hydration (#782)

### LINKED ISSUE #782: Remove "Person has physically left RESET" question from Care exit flow
So that we don't confuse the medical staff with the sequencing of property return and physical exit.

ACs:

- [ ] Remove the "Person has physically left RESET" question
- [ ] Save state in the exit form: when a user has made at least one response, mark form as "edited": button title on previous page should now be "Finish exit" (this behavior already exists but is currently tied to the "Personal has left RESET" question
- [ ] Remove "Save exit details" state on CTA - should now just be "Confirm exit"
- [ ] "Confirm exit" button should be disabled if: exit questions are not all answered (and if person has property, "I've confirmed with Deputy..." question is not answered)

Image

beaudrykock-moi:
## Summary

- Removed the “Person has physically left RESET?” question from the care exit form.
- Changed the exit form CTA to always read “Confirm exit” and removed the “Save exit details” state.
- Updated confirm-exit enablement so all four exit questions must be answered, plus property-handling confirmation when the person has associated property.
- Preserved in-progress exit form selections locally so returning via “Finish exit” restores previous answers.
- Updated the prior-screen “Start exit” / “Finish exit” state to be driven by edited exit form state instead of the removed physical-left question.
- Replaced the property-handling yes/no control with a Figma-aligned checkbox and moved visual overrides into a CSS module to avoid inline styling.

Closes #782


beaudrykock-moi:
@francisli as far as I could tell, the Figma design for the check box was not implementable with the available themes, hence the custom implementation. Don't have a strong opinion on this, so AOK to go with theme if you think that's just better.

francisli:
> @francisli as far as I could tell, the Figma design for the check box was not implementable with the available themes, hence the custom implementation. Don't have a strong opinion on this, so AOK to go with theme if you think that's just better.

You're correct, the Figma design modifies the appearance from the default. My change just moves your custom implementation out of that specific modal into our customized AppTheme so that it will be easily re-usable elsewhere.

beaudrykock-moi:
Sorry @francisli! Just read the commit, didn't look at the details. Appreciate you taking care of that.

francisli (approved):
I moved the Checkbox appearance customization out of the modal into our custom AppTheme so that it can be used elsewhere.

Future work- we should really just persist "draft" exit survey responses to the server rather than use local storage. The staff are using a fleet of shared iPads, having the draft stored locally on just one device doesn't do much if someone else is going to finish the exit on another device.

beaudrykock-moi:
Ack, right you are. Created https://github.com/sfbrigade/care-connect/issues/833

PR #805, opened 2026-04-29
Adjusted title, copyright, and URL in email (#734)

### LINKED ISSUE #734: Adjust copy in release and 647f e-mails
The e-mails sent for with the 647f and the release docs include some unnecessary language around copyright and the wrong URL.

AC:

- [ ] Change title to "CCSF CareConnect"
- [ ] Remove the copyright language
- [ ] Correct the URL to "https://careconnect.sf.gov"

Image

Image

beaudrykock-moi:
## Summary

Updates all application-generated emails to use the official CareConnect email branding.

## Changes

- Replaces the email title with `City and County of San Francisco: CareConnect`
- Updates the email footer URL to `https://careconnect.sf.gov`
- Removes copyright text from shared HTML and plain-text email footers
- Adds `EMAIL_SITE_TITLE` and `EMAIL_BASE_URL` environment defaults for email-specific branding
- Updates email tests to assert the new title, URL, and absence of copyright text

Note: I suppose by not piping in the URL as the email title this is a bit more brittle if CareConnect ends up being deployed to other sites/centers, but those changes would be so far off and also may not yield ideal text for the copy, so hopefully this approach is acceptable.

Closes #734


francisli (approved):
@beaudrykock-moi I can't see a case when BASE_URL and EMAIL_BASE_URL will be different but I'll leave it in there...

Just FYI, if you received emails of forms with "https://careconnectsf.org" in the footer, that is CORRECT- recall we had URL mixups with some people still using the DEV environment, so some emails were getting sent from there.

Emails from the prod environment already have https://careconnect.sf.gov as the BASE_URL and in the footer.

PR #810, opened 2026-04-30
FIELD: Improve photo preview state

### LINKED ISSUE #792: Make "Is photo clear enough to use?" larger

saadabdali:
Closes #792

### In this PR:
- Enlarges the font for the `Is the photo clear enough to use?` message (shown after you snap an ID photo)
- During that preview state, use a colored border around the preview photo (to hint at the mode transition from live camera view to photo preview; this pattern is used in many other mobile apps)

### Possible future edits:
I wanted to keep this change small, but I might also suggest:
- Reword the message. `is the photo clear enough?` might make the user feel bad, like we're telling them there's a problem with the photo. Could make it an action instead of a question: `Review your photo`.
- Move the message up to be shown in the same spot as the `Place the ID in the frame` message on the previous screen

## Before and after
image


saadabdali:
Tagging @francisli for code review; tagging @uxuichelsea because I think she originally pointed out the problem to me during the simulation at RESET.

francisli (approved):

PR #818, opened 2026-04-30
Pass --skip-generate to prisma db push in test helper

grahamplace:
## Summary

Fixes flaky `Cannot find module '#main-entry-point'` failures in CI under the parallel server test runner.

## Root cause

`server/test/helper.js` runs `npx prisma db push` per test file. By default, `prisma db push` *also regenerates the Prisma client* into `node_modules/.prisma/client/`. With `TEST_CONCURRENCY=8`, up to 8 test processes simultaneously rewrite the same `node_modules/.prisma/client/package.json`. If a sibling process happens to load Prisma during the window where one of those rewrites has the file open with the `imports` field not yet written, Node's CJS resolver fails:

\`\`\`
Cannot find module '#main-entry-point'
Require stack:
  /__w/care-connect/care-connect/node_modules/.prisma/client/default.js
\`\`\`

`default.js` does `require('#main-entry-point')`, which is a Prisma package-internal subpath import that resolves via the imports field of `node_modules/.prisma/client/package.json` — the exact file being concurrently rewritten.

This is distinct from the `--experimental-test-module-mocks` loader race that #732 already fixed via the two-pass split. That race is per-process (the experimental flag patches the CJS resolver in-process); this one is filesystem-level across processes.

## Fix

Pass `--skip-generate` to `prisma db push`. The `pretest` script already runs `prisma generate` once before the test suite starts, so the client is up to date and there's no reason to regenerate it inside every test process.

## Test plan

- [ ] CI passes on this PR
- [ ] No `#main-entry-point` errors in the next several CI runs across other PRs that branched off latest dev


saadabdali (approved):
I'm no expert in prisma, but this looks reasonable to me.

PR #822, opened 2026-04-30
Adds org admin ability to edit person details (#762)

### LINKED ISSUE #762: Allow editing of a user's name by an admin
So that mistakes can be corrected

AC:

- [ ] Admin user can edit the name of any user

beaudrykock-moi:
## Summary

- Added org-admin managed user detail and edit routes under `/manage-users`.
- Made active user cards open a profile-style details view with personal information and SFPD/SFSO position details.
- Added personal information and position details edit screens with disabled save buttons until values change.
- Shows a success toast and returns to the details view after saving changes.
- Updated active user cards to better match the Figma feed card typography, spacing, colors, and button sizing.
- Expanded org-admin API permissions so org admins can view and update allowed profile fields for users in their own organization.
- Added row locking for user updates via a Prisma `findByIdForUpdate` helper and re-checks org scope under lock.

Also:
- Encountered a testing error and increased the facility-status transaction timeout to 30s so expected lock waits during concurrent facility-close/hold-create races do not expire the transaction and return a 500.

Closes #762


saadabdali (approved):
Approved, with a few notes.

- I don't love raising the transaction timeout in `status.js` from its default time (I _think_ 5s?) up to 30s. But I'm not sure how to reason about it. It appears we potentially do a lot of work under a lock when closing a facility. Maybe the real solution is to refactor that. Am leaving this for now because I don't know whether it's worse to leave the default timeout (risking timeouts/failures in production) or raising it (which could create 30-second periods where nobody can obtain a lock or edit anything.)
- I see that the designs called for letting users edit a `Full Name` field instead of `First Name` and `Last Name` fields. Presumably this was done to make the form appear lighter. But the underlying DB stores first name and last name as separate fields, and every other name form shows them that way. I pretty strongly dislike this disconnect, because the user cannot actually see how the name he types in will get parsed and stored. A user could edit a name merely by loading and saving this form (`Mary Jane` `Watson` would silently get re-saved as `Mary` `Jane Watson`). 

Changes I made:
- Refactored the transaction in patch.js to (hopefully) be simpler
- Added a few tests to cover the new pages

beaudrykock-moi:
Oh, yeah, agreed, that first name/last name thing is a problem. Great call. I've asked Chelsea to update the design and will implement the first name / last name adjustment.

beaudrykock-moi:
@saadabdali I added the text field adjustment. Would be grateful for one more review from you before merging - and thank you again.

saadabdali:
Taking a look now

saadabdali:
This looks good to me.

francisli (approved):

PR #808, opened 2026-04-29
change pdf generation so editable, email to officer

aaronhans:
still testing locally

will close #783

PR #823, opened 2026-04-30
Officers can certify the info they enter (#767)

### LINKED ISSUE #767: As an arresting officer, I can certify information is true and correct
So that we comply with City Attorney guidance.

AC:

- [ ] Somewhere in the arresting officer flow, add the following language:

"I declare under penalty of perjury that the information I have entered is true and correct, based on my personal knowledge, or is based on information and belief following an investigation of the events and parties involved."

The user must be able to check/select this to indicate their agreement.

- [ ] Include this statement in the 647f at bottom after all fields:

"Officer/Deputy [name] declared under penalty of perjury that the information they entered is true and correct, based on their personal knowledge, or was based on information and belief following an investigation of the events and parties involved."

beaudrykock-moi:
## Summary

Adds officer certification to the arrest/details flow before transfer.

Officers now move from the personal property form to a new “Certify information” screen, where they must read and agree to the declaration before finishing details. Certification state is saved immediately so progress is preserved if the user exits the flow, and completing the flow returns the officer to the home screen with the existing saved toast behavior.

## Changes

- Adds a new `CertifyInformation` view and `/holds/:id/certify` route.
- Updates the personal property form CTA to `Next: Certify`. (Slight deviation from spec, to keep things shorter.)
- Adds `certifiedAt` to the deflection/person details data model.
- Saves `certifiedAt` when the declaration checkbox is checked, and clears it when unchecked.
- Enables `Finish details` only after certification is checked.
- Adds certification to client and server completeness validation.
- Blocks transfer/arrival gating until incident details, person details, and certification are complete.
- Shows certification status in the deflection incident overview:
  - `Certified as true and correct at [time] on [date]`
  - `Incomplete` when not yet certified
- Adds focused tests for certification UI behavior, validation/completeness, and updated transfer/arrival expectations.
- Updates affected route tests and fixtures so flows that require complete details now include certification explicitly.

## Notes

- Did not implement the second AC around 647f generation. Aaron's #808 has changed PDF gen so approach will be implemented separately.

Closes #767


saadabdali (approved):
I made two under-the-hood changes which I believe are invisible to the user:
- The original PR used a raw `<Box>` and styled it; I replaced this with a Mantine `<Checkbox>` and ensured that it looks the same. It saves us some custom CSS.
- The test code had a helper function duplicated in 3 places; unified that.

Each of these is in its own commit and can be undone if we don't like it.

Couple other things I noticed:
- The certification form is the only page in the Hold flow that has an X in the top-right corner. It would probably be better to either add it to every form page, or remove it here, but that's not a merge-blocking concern. Users probably won't notice.
- In this PR as written, we allow the arresting officer to hand off a Hold in an un-certified state. In that situation, the app will force the officer who receives the Handoff to click the certification before custody transfer. 
- In this PR as written, an officer who receives a Handoff can un-certify and re-certify the Hold. If the user does this, it will update the `certifiedAt` timestamp (which will only track the most recent certification). 

I'm not sure if any of those are blockers - that's really a policy question.

beaudrykock-moi:
@saadabdali added the gating code and close button. Grateful for one more review pass before merge - thank you!

saadabdali:
This looks good to me.

Just noting for future:
- If I have three holds and want to hand off one of them, then the app will now require me to complete deflection details on all three before I can reach the Handoff screen (to hand off the one). This is not ideal, but it's only an inconvenience (it's requiring the user to do more than we strictly need from them; and maybe it's an uncommon situation.)

I'd vote to merge this as-is, and then in future we can reconsider the Handoff UX. One possible direction:
- "Initiate a Handoff" becomes a global action instead of a per-Incident action
- the subsequent Handoff screen shows you all of your holds
- Any hold that's not ready for Handoff shows a locked QR code and communicates why that Hold is ineligible (incident details incomplete, or deflection details incomplete)

PR #826, opened 2026-04-30
Tag PostHog events with git SHA + upload client source maps

grahamplace:
Two related observability improvements before going to prod:

1. **Release tagging.** Every PostHog event (client + server) now carries `$git_commit` and `release` properties set to the commit SHA. PostHog Error Tracking will group exceptions by release, so you can attribute spikes to specific deploys.

2. **Source maps.** Client Vite build now emits `'hidden'` source maps (no `sourceMappingURL` in shipped JS, so browsers don't fetch them). The Dockerfile uses `npx @posthog/cli sourcemap inject` + `upload` to send maps to PostHog at build time, then deletes the `.map` files from the image. Without this, prod stack traces are minified and unreadable.

## Required to fully enable

Two repo-level configs are needed; the build still succeeds without them (skip message printed, source maps still stripped from the image).

1. **Repo secret `POSTHOG_CLI_API_KEY`** — personal API key. Recommended setup: organization-scoped key with these permissions:
   - `Error tracking → write`
   - `Organization → read`

   One key uploads to all projects, since the same Docker image is built once and deployed to multiple environments.

2. **Repo variable `POSTHOG_CLI_PROJECT_IDS`** — space-separated list of PostHog project IDs (one per environment). The Dockerfile loops `sourcemap upload` over each. Verified locally that maps land in both dev and prod projects under the same release.

## Notes

- `GIT_SHA` flows in via `--build-arg` from CI (`github.sha`) → Dockerfile `ARG` → `ENV VITE_GIT_SHA`. Server reads it at runtime via `process.env.VITE_GIT_SHA`; client reads it at build time via `import.meta.env.VITE_GIT_SHA`.
- Source maps are generated for the client bundle only; SSR build doesn't need them.
- `posthog-cli` version pinned (`@0.7.11`) so a transitive update can't surprise CI.
- `inject` only needs to run once (it talks to whichever project is set first); `upload` is the per-project step. Each project's release is auto-created on first upload.


Dockerfile
```diff
@@ -34,5 +40,19 @@ RUN npm install && \
...
 
+# Upload client source maps to PostHog for stack-trace symbolication, then delete
+# them from the image so they aren't shipped to clients. Skipped if the
+# POSTHOG_CLI_TOKEN secret isn't present (e.g. local builds, forks).
+RUN --mount=type=secret,id=posthog_cli_token \
+    if [ -s /run/secrets/posthog_cli_token ]; then \
+      POSTHOG_CLI_TOKEN="$(cat /run/secrets/posthog_cli_token)" \
+        npx --yes @posthog/cli@0.7.11 sourcemap inject --directory ./client/dist/client && \
+      POSTHOG_CLI_TOKEN="$(cat /run/secrets/posthog_cli_token)" \
+        npx --yes @posthog/cli@0.7.11 sourcemap upload --directory ./client/dist/client; \
```
fwextensions: Curious how the git sha gets associated with the sourcemap.  Is the posthog cli looking for that env var?

grahamplace: I pushed a change to explicitly pass `--release-version"$GIT_SHA"` to make it clearer, before the posthog cli derives it from git if CLI flags aren't specified

francisli:
@grahamplace have you tried running the posthog cli commands manually? I'm reviewing the corresponding docs at:

https://posthog.com/docs/error-tracking/upload-source-maps/cli

And it seems like there needs to be a "login" step first that also needs POSTHOG_CLI_PROJECT_ID env var set to know which project to upload the maps to.

If that's the case, don't we need to repeat and loop over all the different project environments (dev, stage, prod) with their corresponding IDs...?

grahamplace:
Great catch @francisli, I misunderstood how this worked. I think you were right on both counts. Pushed a change to do what you proposed

Now we set `POSTHOG_CLI_PROJECT_IDS` (space-separated) so to loop and upload maps to each environment's PostHog project

I ran inject/upload via CLI for project ids: 253384 & 253393, you can see the uploaded source maps:
- https://us.posthog.com/project/253384/error_tracking?activeTab=configuration#selectedSetting=error-tracking-symbol-sets
- https://us.posthog.com/project/253393/error_tracking?activeTab=configuration#selectedSetting=error-tracking-symbol-sets

PR #827, opened 2026-04-30
Filter known-noise exceptions before PostHog ingest

grahamplace:
Adds a `before_send` hook to `posthog-js` init that drops three categories of known-useless exceptions before they reach Error Tracking:

- **`ResizeObserver loop` warnings** — browser-internal, never actionable.
- **`Script error.` / `Non-Error promise rejection captured`** — cross-origin script errors the browser intentionally redacts; nothing we can debug.
- **Errors originating from `chrome-extension://` / `moz-extension://` / `safari-web-extension://` / `webkit-masked-url://`** — user-installed extensions throwing in our page context.

Helper is module-scoped; can be expanded by adding to `NOISE_PATTERNS` or `EXTENSION_FILENAME` as more noise shows up in the dashboard.

### DIFF
--- a/client/src/analytics/PosthogProvider.jsx
+++ b/client/src/analytics/PosthogProvider.jsx
@@ -5,6 +5,26 @@ import { useStaticContext } from '@/StaticContext';
 
 const DEFAULT_API_HOST = 'https://app.posthog.com';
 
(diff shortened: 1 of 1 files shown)


grahamplace:
Folded into #820 — noise filter belongs with the change that turns on client-side exception capture in the first place.

PR #841, opened 2026-05-01
Adjusts email template to surface CareConnect (#840)

### LINKED ISSUE #840: Adjust email title so it doesn't run so long and is clearer in inbox
CareConnect is no longer easily visible in the inbox, because the "City and County of San Francisco" obscures it.

Image

Update so that "CareConnect" is the primary title in the e-mail and inbox.

beaudrykock-moi:
## Summary

Split email branding into separate concerns so inbox presentation can stay short while the email body retains the full formal title.

### Changes
- Added `EMAIL_FROM_NAME` for the SMTP `From` and `Reply-To` display name
- Added `EMAIL_SUBJECT_TITLE` for subject-line branding
- Kept `EMAIL_SITE_TITLE` as the in-email/body header title
- Updated mailer fallbacks so existing deployments continue to work if the new env vars are unset
- Updated subject templates for verification code, password reset, and invite emails to use `EMAIL_SUBJECT_TITLE`
- Added the new env vars to `server/example.env` and ECS deployment wiring

### Result
Emails can now render as:
- From: `CareConnect <no-reply@...>`
- Subject: `Your CareConnect verification code`
- Body header: `City and County of San Francisco: CareConnect`

Closes #840


saadabdali (approved):
Looks good to me. 

**Deployment note**: When shipping this code to a new environment, we will also need to set two new env vars in order for the new behavior to take effect. (They are mentioned in the updated `example.env` in this PR)

```
EMAIL_FROM_NAME="CareConnect"
EMAIL_SUBJECT_TITLE="CareConnect"
```

francisli (approved):

PR #848, opened 2026-05-03
Makes 849b org-generic (#847)

### LINKED ISSUE #847: Language inserted into 849b needs to be org-aware
In the 849b, the following language is inserted:

"The SFPD Officer who brought the person to RESET recorded the following observations on the 647(f) documentation"

Now that both SFPD and SFSO Officers can do arrests, this needs to be changed to be more generic.

"The law enforcement officer who..."

beaudrykock-moi:
## Summary
- update the 849(b) release narrative copy to remove the hard-coded `SFPD` reference
- apply the same wording change in both the client-side default narrative builder and the server-side PDF narrative builder
- update affected tests to match the new narrative text

Closes #847


aaronhans (approved):
text change only, looks good

Issue #852, opened 2026-05-03
Facility admin sees option to cancel expired hold


beaudrykock-moi:
Actual:
1. Facility_admin still sees the "cancel hold" button on an expired hold

Expected:
1. Expired holds should not be cancelable - they are already 'canceled'.

<img width="270" height="135" alt="Image" src="https://github.com/user-attachments/assets/9658b778-81a7-4f99-be67-6e83a7492d63" />

Issue #851, opened 2026-05-03
Facility admin cannot cancel active hold


beaudrykock-moi:
Attempted as FACILITY_ADMIN to cancel a hold from the Manage Capacity flow and received a 403 from DELETE /api/deflections/:id. Both for an active and an expired hold (tested in prod).

Steps to reproduce:

1. Sign in as a user with the FACILITY_ADMIN role.
2. Navigate to /manage-capacity.
3. Open “Manage chair holds.”
4. Select an active.
5. Click “Cancel hold.”

Actual result:

1. Frontend sends DELETE /api/deflections/:id
2. Backend responds with 403
3. User sees a failed cancel action

Expected result:

1. Cancel hold action by facility_admin should succeed for an active hold

PR #858, opened 2026-05-04
Fixes in-transit list issues (#857, #851, #852, #853)

### LINKED ISSUE #857: deflection.expiry() is not touching awaiting-transfer states
This can lead to situations where expired holds in awaiting-transfer states lead to "in transit" counts that can't be normalized.

beaudrykock-moi:
## Summary

This PR fixes several hold-management regressions and inconsistencies across facility-admin and field-user workflows.

### Facility-admin hold cancellation
- Fixed a `403` when a facility admin cancels an active hold.
- Scoped the permission change to the cancel route only, without broadening general deflection-edit permissions.
- Added a regression test for facility-admin hold cancellation.

### Facility-admin "Holds in transit" panel
- Fixed the manage-capacity hold list to show only active in-transit holds.
- Updated the query to request only `ACTIVE` + `DETAINED` holds.
- Added a defensive client-side filter so cancelled, expired, and onsite-awaiting-transfer holds do not render in that panel.
- Added focused client regression coverage for the request shape and rendered list.

### Facility-admin "Holds awaiting custody transfer" panel
- Added awaiting transfer lists
- Moved awaiting transfer and in transit into accordion

### In-transit boundary normalization
- Changed the in-transit definition so it now includes only active holds in `DETAINED`.
- Moved the counter boundary from transfer-time to arrival-time:
  - `/facilities/:facilityId/arrived` now decrements `bedType.inTransit` when a hold transitions from `DETAINED` to `ONSITE_AWAITING_TRANSFER`
  - `/deflections/:id/transfer` no longer decrements `inTransit`
  - cancel/reopen logic now affects `inTransit` only for `DETAINED` holds
- Updated related server tests to match the new boundary.

### Bed-type API counter correctness
- Fixed facility bed-type read endpoints to derive `inTransit` from live deflection rows instead of trusting a possibly stale stored column. 
- `inTransit` is now computed from active `DETAINED` deflections on read.
- Added a regression test proving the API returns the derived value even if the stored `BedType.inTransit` field is stale.

This does not necessarily solve the problem of the past drift in the stale column, but making more aggressive changes seemed risky at this stage.

## Why

Before these changes:
- facility admins could not cancel holds they were expected to manage
- the manage-capacity UI could show cancelled/expired/on-site holds in the in-transit list
- the `inTransit` concept was inconsistent across UI, counters, and mutation flows
- the bed-type API could expose stale denormalized `inTransit` values

After these changes:
- facility-admin hold management matches intended permissions
- "Holds in transit" reflects only true in-transit holds
- the arrival boundary and the displayed counter semantics are aligned
- bed-type API consumers get a correct `inTransit` count even if stored counters drift

Closes #857, #851, #852, #853

### DIFF
--- a/client/src/lesc/components/ManageCapacity/ManageHolds.jsx
+++ b/client/src/lesc/components/ManageCapacity/ManageHolds.jsx
@@ -1,13 +1,23 @@
 import { useState } from 'react';
 import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query';
-import { Button, Card, Group, Loader, Modal, Stack, Text, Title } from '@mantine/core';
+import { Accordion, Button, Card, Group, Loader, Modal, Stack, Text, Title } from '@mantine/core';
+import { DateTime } from 'luxon';
 import { Link } from 'react-router';
 
 import Api from '@/Api';
 import { useToast } from '@/components/ToastContext';
+import { facilityLiveQueryOptions } from '@/hooks/facilityLiveQueryOptions';
 import useNow from '@/hooks/useNow';
 import useSubjectDetails from '@/hooks/useSubjectDetails';
 import { formatTimeRemaining } from '@/utils/format';
+
+function isCurrentlyActiveHold (hold, now) {
+  if (!hold || hold.status !== 'ACTIVE') return false;
+  if (!hold.expiresAt) return true;
+
+  const expiresAt = DateTime.fromISO(hold.expiresAt);
+  return !expiresAt.isValid || expiresAt >= now;
+}
 
 function HoldCard ({ deflection, facilityName, onCancelClick }) {
   const now = useNow(1000, deflection.status === 'ACTIVE' && !!deflection.expiresAt);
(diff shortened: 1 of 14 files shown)


client/src/lesc/components/ManageCapacity/ManageHolds.jsx
```diff
@@ -106,11 +111,11 @@ function ManageHolds ({ facility }) {
       <Stack>
         <Title order={4}>Holds in transit</Title>
 
-        {holds?.length === 0 && (
+        {activeHolds.length === 0 && (
```
aaronhans: adding !isLoading && to this condition will prevent possible loading flash of this "No holds" message.

aaronhans (changes requested):
I think there might be a role issue here. The facility admin in the test data also has the CARE role. In production beaudry's email has roles : ORG_ADMIN,FACILITY_ADMIN,FIELD. He doesn't have the CARE role. I think the lack of that role would still prevent you from seeing the 3 rogue records.

beaudrykock-moi:
> I think there might be a role issue here. The facility admin in the test data also has the CARE role. In production beaudry's email has roles : ORG_ADMIN,FACILITY_ADMIN,FIELD. He doesn't have the CARE role. I think the lack of that role would still prevent you from seeing the 3 rogue records.

I had the CARE role previously but removed it.

client/src/lesc/components/ManageCapacity/ManageHolds.jsx
beaudrykock-moi: Adding now

aaronhans:
I think we need to expand permissions here so that people who have org and facility admin roles can see everything for the facility even if they aren't a super admin

https://github.com/sfbrigade/care-connect/blob/dev/server/routes/api/deflections/list.js#L127

beaudrykock-moi:
> I think we need to expand permissions here so that people who have org and facility admin roles can see everything for the facility even if they aren't a super admin
> 
> https://github.com/sfbrigade/care-connect/blob/dev/server/routes/api/deflections/list.js#L127

Not sure why? Doesn't this scope in admins and facility-scoped care/custody users? Those are the folks who need to see the full list.

francisli:
> > I think we need to expand permissions here so that people who have org and facility admin roles can see everything for the facility even if they aren't a super admin
> > https://github.com/sfbrigade/care-connect/blob/dev/server/routes/api/deflections/list.js#L127
> 
> Not sure why? Doesn't this scope in admins and facility-scoped care/custody users? Those are the folks who need to see the full list.

We need to clean up this endpoint, but this logic is correct- care and custody users already can see all deflections in a facility. I'm going to merge this in now, so that the offending deflections can be cancelled to reset the availability back to the full 25.

francisli (approved):

PR #872, opened 2026-05-04
Gives release unit precedence (#869)

beaudrykock-moi:
## Summary

- Update Certificate of Release unit selection to use the releasing deputy’s unit rather than the incident unit.
- Remove fallback to the originating incident/SFPD unit for the Certificate of Release - blank is better than wrong.
- Add form unit tests covering the SFSO unit case and blank-unit fallback.

Closes #869


aaronhans (approved):
updated logic for unit identifier

PR #873, opened 2026-05-04
tighter capacity count

aaronhans:
We are seeing old test data influence the capacity counts returned in the capacity api. Tighten up how this is calculated:

- `occupied` counted rows with \`arrivedAt != null AND exitedAt = null\` but had no \`status\` filter, so cancelled-after-arrival and stale pre-go-live test rows were counted forever.
- `inTransit` summed \`BedType.inTransit\`, which has drifted from the actual ACTIVE+DETAINED row count.

Mirror the pattern already in [bed-types/list.js](server/routes/api/facilities/_facilityId/bed-types/list.js): compute both counts live from row state and ignore the denormalized counters. The counters on \`BedType\` are unchanged — other endpoints behave identically.

### DIFF
--- a/server/routes/api/status/capacity.js
+++ b/server/routes/api/status/capacity.js
@@ -1,6 +1,8 @@
 import { StatusCodes } from 'http-status-codes';
 import { z } from 'zod';
 
+import Deflection from '#models/deflection.js';
+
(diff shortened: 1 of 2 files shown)


beaudrykock-moi (approved):
Looks sensible to me.

PR #881, opened 2026-05-04
Fix unhandled rejection in QRScanner unmount cleanup

grahamplace:
## Summary

`Html5Qrcode.clear()` returns `void` (synchronous), not a `Promise` — confirmed in `node_modules/html5-qrcode/html5-qrcode.d.ts:47`. The unmount cleanup in `QRScanner.jsx` chained `.catch(() => {})` on that void return, which threw `TypeError: undefined is not an object (evaluating 'scanner.clear().catch')` synchronously inside the `.finally` callback. That made the outer promise reject with no handler → unhandled promise rejection → captured by PostHog as `TypeError: undefined is not an object (evaluating 'K.clear().catch')` on `/custody` and `/care`.

The other two `clear()` call sites in the same effect already use the right pattern (`try { scanner.clear(); } catch {}`) — line 317 was the holdout. Pre-existing bug; only became visible after PostHog `capture_exceptions` was enabled.

## Test plan

- [ ] Open `ScanCodeModal` on `/custody`, scan or close, confirm no console errors
- [ ] Same on `/care`
- [ ] Confirm `K.clear().catch` errors stop appearing in PostHog after deploy


francisli:
Duplicate of #865

PR #883, opened 2026-05-04
Fixes blank printing issue and enlarges time fields (#878)

### LINKED ISSUE #878: Certificate of release shows content in PDF but prints blank

beaudrykock-moi:
## Summary

- Regenerate Certificate of Release PDF field appearance streams so filled fields print correctly while remaining fillable.
- Remove reliance on `NeedAppearances`, which can display fields digitally but print blank in some PDF viewers.
- Add a repeatable script to repo to widen the Certificate of Release arrest/release time fields in the template PDF.
- Update the certificate template via script so `Time` and `Time_2` have enough width for `HH:MM` values.

Closes #878


aaronhans:
Ah font issues.

Are the package-lock changes intentional?

aaronhans (changes requested):
Looks good but I think we can remove package-lock from the PR

beaudrykock-moi:
Aw man thought I caught that. No, def don't need the package-lock. Will remove that.

aaronhans (approved):

server/lib/forms/cert/fillCoR.js
```diff
@@ -88,9 +88,12 @@ export async function fillCoR (pdfBytes, data) {
     });
   }
 
-  // ── Preserve the form's native fonts: let the viewer render appearances ──
+  // ── Generate printable appearances while preserving fillable fields ──
+  const appearanceFont = await pdfDoc.embedFont(StandardFonts.Helvetica);
```
fwextensions, line 92: Interesting.  I wouldn't have thought that Helvetica would be a standard embeddable font, as most OSes don't come with a license for it.  The linux distro the server is running on doesn't even have the "standard" Microsoft fonts.

PR #885, opened 2026-05-05
Updates 647f template with new fields (#879)

### LINKED ISSUE #879: 647f template updates
At SFPD request, 647f should include case number and space for supervising sergeant to add a signature.

AC:

- [ ] Row appears below CAD number, labeled "Case number", with case number inserted
- [ ] Row appears at bottom, for sergeant to sign off on

beaudrykock-moi:
## Update

Added the requested 647f PDF template fields while preserving the existing form field IDs.

### Changes

- Added a new fillable `Case_Number` field directly below the existing `CAD_Number` row.
- Wired `incident.caseNumber` into 647f generation so the case number is populated automatically whenever the form is generated.
- Added fillable fields at the bottom of the PDF for:
  - `Supervising_Sergeant_Name_Star_Number`
  - `Signature`
- Adjusted the page 1 layout below `CAD_Number` so the new row fits cleanly and matches the existing label styling.
- Adjusted the bottom page layout so the supervising sergeant label is no longer overlapped by its fillable field.
- Updated the React 647f rendering path and unit coverage for case number propagation.

Closes #879


aaronhans (approved):
cool, minimal changes adding fields

PR #888, opened 2026-05-05
Surface unhandled rejections and errors in dev console

grahamplace:
When trying to reproduce the error for https://github.com/sfbrigade/care-connect/pull/881, I realized I wasn't actually seeing the unhandled exception log in my browser console which was confusing. I had to do this in the console to see it when reproing:
```
  window.addEventListener('unhandledrejection', e => console.error('UNHANDLED:', e.reason));
```
So this PR aims to surface the errors that Posthog swallows in dev only.

## Summary

PostHog's `capture_exceptions: true` installs `unhandledrejection` and `error` listeners that call `preventDefault()`, which suppresses the browser's default console log for those events. In dev that hides real bugs — the `QRScanner.clear()` unhandled rejection on `/custody` and `/care` had been firing on every modal close without anyone noticing because the browser console was silent.

Add a dev-only pair of listeners at the Vite entry point that re-emit both events via `console.error`. Pure DX, prod is unaffected.

## Design

- **Dev-only** (`import.meta.env.DEV`) — prod still relies on PostHog Error Tracking; we don't want to log to users' consoles.
- **Doesn't preventDefault, doesn't stopPropagation** — PostHog still captures normally, this is purely additive.
- **Runs at entry** — installed before any other module, catches everything from boot.
- **Both events** — `unhandledrejection` is the one PostHog hides most visibly, but it also intercepts `error` (uncaught throws/render errors). Symmetry.
- **Possible duplication** — when PostHog *doesn't* preventDefault, the browser will still log too, so you'd see one line from us plus the browser's own. Tolerable in dev.


aaronhans (approved):
Nice! 

Locally I was still seeing errors but that was because I never set VITE_POSTHOG_KEY in my env file. 

It is good to have the behavior to show errors on dev always in code so people don't get confused if they do set a local env value.

PR #896, opened 2026-05-05
In transit holds view for CUSTODY role (#862)

beaudrykock-moi:
## Summary

Redesigns the CUSTODY/SFSO home page to match the new Figma Transit/Custody/Released flow.

### Details

- Added a new first tab: **Transit**
- Renamed existing tabs:
  - `In custody` → `Custody`
  - `Legally released` → `Released`
- Shows `DETAINED` active holds in the Transit tab
- Groups Transit holds by owning/current officer using first initial + last name + badge number (not in design but per SFPD request)
- Shows `Details incomplete` for Transit records with incomplete incident/deflection details
- Shows a `Details` button only when incident/deflection details are complete
- Routes Transit `Details` to the existing read-only custody detail view
- Adds pre-transfer detail header content:
  - `Expires in ...`
  - `Awaiting arrival` status chip
- Hides edit actions and the `849(b)` narrative section for pre-transfer custody detail views
- Removes segmented-control dividers to match Figma
- Uses Mantine theme tokens/CSS modules for newly added styling
- Exposes `currentOfficer` in the deflection list API response for Transit grouping
- Ensured canceled DETAINED holds do not persist in Transit by forwarding `status: ACTIVE` through the API client

### Tests

- Added/updated focused tests for:
  - Transit grouping by officer
  - Transit request filtering to active DETAINED holds
  - API `status` query param passthrough
  - Pre-transfer read-only details without `849(b)` narrative
  - Pre-transfer detail header status/expiry display

Closes #862


client/src/lesc/components/custody/Custody.jsx
```diff
@@ -21,9 +25,11 @@ import CustodyCard from './CustodyCard';
 
 import ScanTransferCodeModal from './ScanTransferCodeModal';
 import { RELEASE_TOAST_KEY } from './LegalReleaseQuestions';
+import classes from './Custody.module.css';
 
 const IN_CUSTODY_STATUSES = 'AWAITING_INTAKE,FAILED_INTAKE,READY_FOR_INTAKE,IN_MEDICAL_INTAKE,IN_CHAIR';
 const RELEASED_STATUSES = 'RELEASED,EXITED';
+const TRANSIT_STATUSES = 'DETAINED';
```
aaronhans: It is intentional that ONSITE_AWAITING_TRANSFER is not in these lists right?

This is the state a person is in when an officer clicks "I've arrived". That state should be very brief if everything is going smoothly and the QR code for that individual about to be handed off is visible on that officer's screen. But an accidental "I've arrived" click could happen in the field and then nobody could see that person on these views.

saadabdali: Hm. If a Hold can enter an intermediate state where it won't show up in any of these tabs, that seems wrong to me. And the `CustodyDetailContent` page groups `DETAINED, ONSITE_AWAITING_TRANSFER` together and thinks of them as `pre-transfer` states for the purpose of deciding whether to render the read-only version of that page. I'll take a closer look at this, but I think this is a good catch.

saadabdali: I would have assumed that after the officers taps I've Arrived, the hold should appear in `Custody` (rather than `Transit`), presumably under `Pending Safety Check`

saadabdali:
I'm making an under-the-hood change:

- Currently this PR modifies `list.js` to always attach `currentOfficer` information to every deflection in every response
- `list.js` is used widely in the app, and none of the other invocations (aside from the new `Transit` tab) presently use that info
- so to reduce payload size and limit the amount of officer PII we send over the wire (to only what the app uses), I'm adding a flag in `list.js` so that only invocations that request this information will actually receive it in the response

client/src/lesc/components/custody/Custody.jsx
saadabdali: Aaron, you're right. Currently the person will just disappear from the Transit view after the officer taps `I've Arrived`. They don't appear under the `Pending Safety Check` until after the Custody transfer. I'll tap Beaudry in Slack. I think this is easy to fix if we just decide how to display it.

beaudrykock-moi: Really good catch on this @aaronhans and @saadabdali. That state keeps falling through the cracks. While "Transit" probably isn't the right place for someone awaiting transfer, I for sure don't want 4 tabs, so I think it's fine if we just roll that state into the Transit tab.

beaudrykock-moi: I'll push the fix.

beaudrykock-moi:
OK pushed the expanded status inclusion fix @saadabdali @aaronhans thanks for your review!

aaronhans (approved):
I am happy, now that onsite awaiting transfer is included

saadabdali:
Made a few small additional [changes after discussing in Slack](https://sfbrigade.slack.com/archives/C09PBES8WSU/p1778013408574419?thread_ts=1778009773.228459&cid=C09PBES8WSU):
- (Beaudry's change above) holds that are in `ONSITE_AWAITING_TRANSFER` are included in the `Transit` view
- For holds that are ONSITE_AWAITING_TRANSFER, don't show the Expiration clock (since the system as a whole disregards the timer after arrival)
- For holds that are ONSITE_AWAITING_TRANSFER, show an `Arrived at x:xx AM` label, and float those groups to the top of the screen for visibility.

image

saadabdali (approved):

PR #897, opened 2026-05-05
849b stingy regen

aaronhans:
## Problem

We were surprised when we attempted to retrieve an already generated 849 by visiting the api route with the id and it regenerated the form using the currently logged in user info. The form should not regenerate unless data changes. Other forms like 647 already work this way, no regeneration unless data changes.

## Testing

Before the fix we could go to the pdf route with a hold id when logged in as somebody other than the person who initiated the arrest and it would put your currently logged in account as the officer doing the arrest.

sample url: ```http://localhost:3333/api/forms/849b/pdf/120```

After the fix the pdf is not regenerated by visiting the API route with the hold id so the arresting officer stays static.

### DIFF
--- a/server/jobs/generateForms.js
+++ b/server/jobs/generateForms.js
@@ -11,8 +11,6 @@ export default async function generateForms (data, prismaClient = prisma) {
     ? Object.fromEntries(Object.entries(FORMS).filter(([id]) => formIds.includes(id)))
     : FORMS;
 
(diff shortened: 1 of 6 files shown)


aaronhans:
Closing as duplicate of PR #901

beaudrykock-moi:
Sorry @aaronhans didn't see this fix was already up!

PR #903, opened 2026-05-05
Fix/898 fixes to 849b fields

beaudrykock-moi:
## Summary

- Populate additional 849(b) PDF fields during generation (at Deputy's request):
  - Disposition code: `DET/REL`
  - Reporting party code: `R1`
  - Reporting party name from the releasing deputy: `Last, FirstInitial, #Star`
  - Reporting party phone: `415-575-6461`
  - Reporting party business address: `70 Oak Grove St`
  - Reporting party ZIP: `94107`
- Fix the stray `2` appearing near the Citation / Booking Approved area by no longer writing to the unrelated `Text4` field.
- Keep the report page count field populated correctly as `2`.
- Source Prop 115 certification from the original releasing user, not the user who happens to generate the PDF.
- Preserve first-initial deputy formatting after the dev merge.

Closes #898

### DIFF
--- a/client/src/lesc/components/care/CareExitDetails.jsx
+++ b/client/src/lesc/components/care/CareExitDetails.jsx
@@ -47,7 +47,11 @@ function CareExitDetails () {
   const fromDetail = searchParams.get('from') === 'detail';
   const backTo = getCareExitBackTo({ fromDetail, id });
 
+  // Mirrors CARE_EXIT_DESTINATIONS in server/lib/careExitDestinations.js.
+  // Server rejects JAIL on /exit and /exit-details; this filter keeps the
+  // UI honest so users don't pick a destination that will 400.
   const exitDestinations = Object.entries(t('deflectionExitDestination', { returnObjects: true }))
+    .filter(([id]) => id !== 'JAIL')
     .map(([id, name]) => ({ id, name }));
   const exitHousingStatuses = Object.entries(t('deflectionExitHousingStatus', { returnObjects: true }))
     .map(([id, name]) => ({ id, name }));
--- a/client/src/lesc/components/custody/CustodyDetailContent.jsx
+++ b/client/src/lesc/components/custody/CustodyDetailContent.jsx
@@ -215,7 +215,6 @@ function CustodyDetailContent ({ deflection, backTo = '/custody', viewerMode = '
     incidentReady: !deflection?.incidentId || incidentQuery.isFetched,
   });
 
-  const doc849b = deflection?.deflectionDocuments?.find(d => d.formId === '849b');
   const docCert = deflection?.deflectionDocuments?.find(d => d.formId === 'cert');
   const showPostReleaseNarrativeActions = !isCareView && isCustody && isPostRelease;
 
@@ -230,8 +229,7 @@ function CustodyDetailContent ({ deflection, backTo = '/custody', viewerMode = '
   });
 
   function open849bPdf () {
-    const url = doc849b?.fileUrl || `/api/forms/849b/pdf/${deflection.id}`;
-    window.open(url, '_blank');
+    window.open(`/api/forms/849b/pdf/${deflection.id}`, '_blank');
   }
 
   return (
--- a/e2e/849b-pdf-fields.spec.js
+++ b/e2e/849b-pdf-fields.spec.js
@@ -110,6 +110,11 @@ test.describe('849(b) PDF field verification', () => {
     expect(dropdown.getSelected()).toContain('Same/On View');
   });
 
+  test('disposition code is detained and released', () => {
+    const dropdown = pdfForm.getDropdown('Dropdown4');
+    expect(dropdown.getSelected()).toContain('DET/REL');
+  });
+
   // ── Subject fields ──
 
   test('subject code is D', () => {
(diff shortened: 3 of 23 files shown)


saadabdali:
Am also going to take a look at this, mostly to see if I learn anything relevant to #875 which I'm working on today.

saadabdali:
Adding some more context from Beaudry:

> This ended up being a lot more involved, because I wanted to standardize 849(b) PDF generation and release/exit email behavior so every emailed or viewed 849(b) was fresh with all the fields - but obviously still storing the generated PDF for later - and there are a lot of routes to do this on. There were also some resulting issues where 647f emails were being erroneously re-sent, so I had to rebuild the "resend the 647f" triggers.
> 
> Some things to note for this PR:
> 
> 
> - 849(b) field insertion was updated for disposition, reporting party info, Prop 115/Post Training checkbox behavior, report page count, and a stray "2" which was floating around on the PDF.
> - 849(b) emailing now uses live PDF generation instead of relying on previously stored documents.
> - Emailed 849(b) PDFs are stored after send, but storage does not block the email.
> - Release and exit routes now share common helpers for queuing release packets and incident-form emails.
> - Automatic release packets include the expected forms, including 647(f), 849(b), and cert where applicable.
> - Separate 647(f)-only transfer emails are now sent only when the 647(f) was actually updated after custody transfer, such as after relevant PII changes.
> - PDF field insertion now respects field max lengths: seed data was causing failures.

aaronhans (changes requested):
I think the reporting deputy in the generated forms is wrong unless I am misunderstanding this. In my local test the arrest was made by "T. SFPD 1, star #1234". I moved the person through the process and after the sheriff released them the 849 says "Reporting Deputy" is the Sheriff that performed the release(T. SFSO, star #5678), not the arresting officer. I don't see the arresting officer who initially certified the incident report on the generated 849. This is reproducible locally. 

Image

beaudrykock-moi:
> I think the reporting deputy in the generated forms is wrong unless I am misunderstanding this. In my local test the arrest was made by "T. SFPD 1, start #1234". I moved the person through the process and after the sheriff released them the 849 says "Reporting Deputy" is the Sheriff that performed the release(T. SFSO, star #5678), not the arresting officer. I don't see the arresting officer who initially certified the incident report on the generated 849. This is reproducible locally.
> 
> Image

@aaronhans that's expected behavior. The 849b should reference the releasing deputy, not the arresting officer.

aaronhans (approved):
OK, cool. Thanks for clarifying the 849 should reference the releasing officer not the arresting officer. 

I verified locally that the information in the generated pdf remains stable no matter which account is viewing the generated document

saadabdali (commented):
Few notes: 

- it looks like when the user clicks 'email me the 849b', the server tries to do everything synchronously, before returning a response. Other email-generation paths in the codebase use an asynchronous path that has retries built in. Unless there's a reason for this to be synchronous, probably best to re-use the async path. (Am happy to make this change if helpful, already have the right context loaded up.) 
- It looks as if we used to read the stored 849b file in some situations, but now we no longer do (we always re-generate on demand). But we're still storing the document. Is this intentional? (Okay if yes - perhaps we just want to store it for auditing/debugging purposes. But just checking.)

beaudrykock-moi:
> Few notes:
> 
> * it looks like when the user clicks 'email me the 849b', the server tries to do everything synchronously, before returning a response. Other email-generation paths in the codebase use an asynchronous path that has retries built in. Unless there's a reason for this to be synchronous, probably best to re-use the async path. (Am happy to make this change if helpful, already have the right context loaded up.)
> * It looks as if we used to read the stored 849b file in some situations, but now we no longer do (we always re-generate on demand). But we're still storing the document. Is this intentional? (Okay if yes - perhaps we just want to store it for auditing/debugging purposes. But just checking.)

Per our chat in person, @saadabdali, for wider visibility:

- Yes please go ahead with the async generation path fix - thank you!
- Let's keep storage for now. Helpful to have the record.

saadabdali:
Implemented, and tests passing locally. Next I will: bring in changes from dev, then run through a manual test plan to ensure I didn't mess anything up.

saadabdali:
Per an IRL convo with Beaudry: in https://github.com/sfbrigade/care-connect/pull/903/commits/d280b32847f7b24b9517b2ac02d1260f61da6f34 I removed the `Jail` option from the `Legal Release` flow available to `CARE` users. (We realized that CARE users were able to record a Jail exit which triggered an 849b email, and we didn't want any of that to happen. Instead, Beaudry said that Care users should not be able to exit a person to Jail.)

No more `Jail` under **Exit Destinations** in the `/care/{id}/exit` page.
image

saadabdali:
Since this PR has evolved a little over time, just wanted to leave a clear record of the main changes I think are in here now:

### 849(b) field fixes (Closes #898)

- Disposition: `DET/REL`.
- Reporting party: code `R1`, name `Last, F, #Star` from the releasing
  deputy, phone `415-575-6461`, address `70 Oak Grove St`, ZIP `94107`.
- Prop 115 / POST Training checkbox now tracks the *releasing* deputy, not
  whoever happens to be generating the PDF.
- Stray `2` no longer written to the citation area.
- Long values are truncated to PDF field max-length so seed/edge data can't
  break PDF generation.

### Always-live 849(b)
- 849(b) is regenerated from current data on every view and every email
  send. (We still store the PDF after sending, but that's no longer used directly.)

### CARE can no longer exit a person to Jail
- Removed the ability of a CARE user to exit a person to Jail.
- Tightened this in both the frontend (removed Jail option from legal release
  form) and backend.

saadabdali (approved):

PR #907, opened 2026-05-05
Pd api more data

aaronhans:
New fields requested by PD and approved by Sheriff are added. Additional fields that Sheriff team didn't agree to yet like release from facility time are being discussed and may be added in future PRs if we have full agreement.

## Testing

- Set a value for ARRESTS_API_KEY in server/.env local file
- ```curl -s -H "Authorization: Bearer yourvalue" "http://localhost:3000/api/arrests?date=2026-05-05"```

Closes #902

### DIFF
--- a/server/routes/api/arrests.js
+++ b/server/routes/api/arrests.js
@@ -12,7 +12,7 @@ const QuerySchema = z.object({
 });
 
 const ArrestSchema = z.object({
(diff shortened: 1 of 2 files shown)


beaudrykock-moi (approved):
Looks great to me.

PR #908, opened 2026-05-06
[Paused] Custody: Show incident time and address for in-transit holds

### LINKED ISSUE #904: CUSTODY: Add incident location and incident time to Transit view
Earlier today we added a `Transit` tab for Custody users (#862 ), which shows users who are either en route to the facility or have arrived but are awaiting custody transfer.

A few fast follow-ups, to help Custody users understand who's in flight and how far away they are:
- show the Incident address on each In Transit hold
- show the time of the Incident

These two pieces of context would help a Custody user gauge how far away the hold is, and how soon they're likely to arrive.
We could/should also make the address tappable for a quick map view.

saadabdali:
Closes #904.

Intent of this quick work is: show more context to help `CUSTODY` users at the Facility understand who's en route to the facility and how soon to expect people.

(We already have signal that deputies want more context. They've asked for live location of transports in progress, which is hard to deliver. This PR adds some context that's quick/easy to deliver.)

Changes:
- For Custody users, for cards in the `Transit` tab:
  - Before this PR: if the hold is already onsite_awaiting_transfer, show `Arrived X:XX AM`
  - In this PR: if the hold is not yet onsite_awaiting_transfer, show `Detained X:XX AM`.
- Below the person's age and sex, show the incident address (tappable to open a map). 
- Minor: removed the word `at` to preserve space. E.g. `Arrived at 4:32 AM` -> `Arrived 4:32 AM`

Note: It might be easy for a user to misinterpret the address as the hold's _home_ address, rather than the incident/arrest address. I considered a few alternatives but didn't find one that I thought was better overall. I figured users would get accustomed to the behavior and learn how to interpret it.

IMG_0227

### DIFF
--- a/client/src/lesc/components/custody/Custody.jsx
+++ b/client/src/lesc/components/custody/Custody.jsx
@@ -15,7 +15,7 @@ import useSessionState from '@/hooks/useSessionState';
 import { facilityLiveQueryOptions } from '@/hooks/facilityLiveQueryOptions';
 import useNow from '@/hooks/useNow';
 import useSubjectDetails from '@/hooks/useSubjectDetails';
-import { formatSmartDateTime, formatTime, formatTimeRemaining } from '@/utils/format';
+import { formatTime, formatTimeRemaining } from '@/utils/format';
 import { isValidDeflection, isValidIncident } from '@/utils/validators';
 
 import ChairAvailabilityCard from '../ChairAvailabilityCard';
@@ -147,9 +147,6 @@ function TransitCustodyCard ({ deflection, highlighted }) {
   // City omitted on display per product (always SF in current scope); buildAddressQuery
   // backfills "San Francisco, CA" so the underlying map link still resolves correctly.
   const incidentAddressDisplay = [deflection?.incident?.addressLine1, deflection?.incident?.addressLine2].filter(Boolean).join(', ');
-  const incidentTimeDisplay = deflection?.incident?.arrestedAt
-    ? formatSmartDateTime(deflection.incident.arrestedAt)
-    : null;
 
   return (
     <Card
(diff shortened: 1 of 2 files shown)


beaudrykock-moi:
Yeah I think there's a high risk this gets confused for their home address.

What were your alternatives, just curious?

@coilin any thoughts here?

saadabdali:
Alternatives I considered:
- add extra text to clarify that it's the address of the arrest, e.g: `Detained at 343 Liberty St` or `Incident: 343 Liberty St`. But I was hesitant to add so much more text to the card - it seems ugly to show the word `Detained` twice on the card. Also wanted to avoid the word `Arrest` due to the language concern around custodial arrest.
- Instead of a word such as `Detained` or `Incident`, we could come up with an icon that clearly communicates the idea of an arrest, and show that icon next to the address. So for example, you might see `<handcuff icon> 343 Liberty St`. But this small PR seemed like the wrong place to introduce a totally new piece of design language that we don't use elsewhere in the application. I'm not sure it would be self-explanatory.
- If we were showing the incident date and time _together_ on the card, it might seem reasonable to use the word `Detained`. So for instance: `Detained 4:22 AM at 343 Liberty St` or `Detained 4:22 AM • 343 Liberty St`. But I preferred the idea of only showing _one_ time on this card (show _either_ the Arrival time if arrived, _or_ the Detention time, but not both), which necessitated separating the two pieces of information.
- Actually, here's a wild idea. What if instead of displaying the address at all, we _only_ displayed an icon (an icon representing a map, or perhaps a _route_) and then when you tap it, instead of showing a _pin_, it shows _the driving route_ from the Incident address to RESET? This would allow you to eyeball-estimate how far the person has to travel, while pretty clearly communicating that this address is where they're coming from while transiting to RESET.

We could also just drop the address from this PR and leave the time only.

We could also show the screenshot to deputies tomorrow and see whether they'd really like that info surfaced there or not (or if it prompts other ideas from them).

We could also drop this PR altogether! No ego here.

beaudrykock-moi:
Thanks, lots of interesting ideas there. Kinda want to leave this with @coilin and [808](https://github.com/orgs/sfbrigade/projects/20/views/2?filterQuery=location&pane=issue&itemId=183605618&issue=sfbrigade%7Ccare-connect%7C866) for now (first part of that ticket is about the location of the arrest), but let's leave this PR around so we can inform the ultimate decision.

coilin:
@saadabdali, thanks for the ideas — they’re interesting

I’d love for us to discuss thoughts like this before they turn into code or mocks, because it’s genuinely interesting to try to push our possibilities as far as we can.

I’m going to take this task on in the next few days and share some ideas from my side as well 

сс @beaudrykock-moi

PR #909, opened 2026-05-06
[Bugfix] Show 'Handed off' label only for holds that were actually handed off

saadabdali:
Closes #880 

### Changes
- Previously: we inferred the `Handed off` label without directly checking the `Handoff` table. The logic was approximately: if your History call to `list.js` returns a hold, and you're not the current officer on the hold, then you handed it off. This logic was brittle, and turned out to be incorrect for admins (whose History call returns _all_ deflections.)
- In this PR: Only show the `Handed off` label for deflections where we have explicitly seen a `Handoff` row where the current user handed this deflection off.

### Not in this PR:
- This PR **does not** change the _scope_ of the History call for admins. I originally said that we should reduce this so that the History view behaves in a more normal way for Admins (and make up for it by providing the right kind of global view for admins elsewhere.) I still think that's directionally correct, but I realized I could (and should) fix this bug more narrowly. We can treat "query behavior for admin users" as a separate problem for a separate PR.
- Note: until we do change the admin behavior, I expect users with admin privileges may continue to see confusing/misleading things in the History view. I'll file a separate ticket for that.

### DIFF
--- a/client/src/lesc/components/Hold.jsx
+++ b/client/src/lesc/components/Hold.jsx
@@ -75,16 +75,16 @@ function Hold ({ incident, deflection, highlighted, onCancelClick, onDetailsClic
                 <Text size='md' c='red.6'>Canceled after expiry</Text>
               </>
             )}
-            {isHandedOff && (
-              <>
-                <Text size='md' c='gray.5'>•</Text>
-                <Text size='md' c='indigo.6'>Handed off</Text>
-              </>
-            )}
-            {!isHandedOff && isCustodyTransferred && (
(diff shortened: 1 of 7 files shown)


aaronhans (approved):
Great job hunting down this issue. Now when I am logged in as admin I no longer see a ton of "Handed off" labels in the History view.

saadabdali:
Thanks @aaronhans. For completeness, the follow-up issue (not necessarily urgent) is filed as #910. (That issue asks: what need are we addressing by returning every single deflection in the History view for admins? and should we address that need differently?)

PR #911, opened 2026-05-06
Make chair counts easier to understand

saadabdali:
@beaudrykock-moi - I spent some time exploring the counts.

On the one hand: this is a small change that I think is ultimately more logical than what we have today.

On the other hand: I also feel I actually ought to ask more questions about who the audience for this change is, and what context they're operating in. And that there's a lot more we _could_ provide here if we knew more about that.

### What's in this PR
- In the Facility card (for `CUSTODY` and `CARE` users), we replace the `In transit` count with a `held` count.
- So now the card shows three numbers: `Available`, `Held`, and `Occupied`. These three numbers are always guaranteed to add up to your _online capacity_ (I'm thinking of online capacity as total capacity, less any chairs marked `unavailableOccupied` or `unavailableUnoccupied` by an admin). 
- Because 'in transit' is still useful, we show that number on a badge directly on the `Transit` tab.

Here's what that looks like for `CUSTODY`:
IMG_0231

`CARE` will also see the updated Facility card, but they don't have a `Transit` tab, so they don't get that badge.

### Why we might merge this
The mental model here is clearer, I think: 
- the numbers in the **Facility Card** are always chair-centric - they're trying to give you an overall picture of how the capacity is currently allocated. In fact, if you look at the old facility card with fresh eyes, it's actually a little strange - it says `10 chairs available` in huge font, and then below that, it says `2 in transit`. Two chairs in transit?
- The **tabbed lists** below that (`Transit`, `Custody`, `Released`) are workflow-centric, and they cut across the natural chair categories. (`Transit` is just a subset of held chairs; `Custody` includes both held chairs and occupied chairs; and `Released` includes some people who are outside the chair accounting.)

### Why we might want to do something else (instead, or in addition)
- In this PR, `CARE` users can no longer see the number of people in transit from the main page. (They can still see it in the detailed breakdown when they click `Manage capacity`). 
- Adding a count badge on the `Transit` tab header uses more horizontal space, which looks fine on my phones but may exacerbate clipping on phones with enlarged fonts.
- There's a question of whether this actually helps the right people. The current Facility card is confusing, but which roles at RESET actually care about that? If it's `CARE`, don't they already get what they need from the detailed breakdown on the `manage-capacity` page? Instead of trying to cram the chair-accounting onto the page header, should we actually offer a full-page, tablet-optimized (rather than phone-optimized) experience? Instead of coarse buckets (held/occupied/available), is it helpful to show a more complete sequence diagram of _all_ the states? In addition to counts, do we actually want to show the underlying Holds in every category? This all depends on who we're imagining the audience for this is, and what they're trying to do.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -128,3 +128,7 @@ care-connect/
 For production, point a wildcard subdomain to the server
 
 - `*.example.com` → Care Connect app
+
+## Further reading
+
+- [`docs/holds-state-machine.md`](docs/holds-state-machine.md) — deflection lifecycle, `BedType` counter invariants, and how the chair-state UI maps onto subject statuses. Read before changing a lifecycle transition or any view that summarizes capacity.
--- a/client/src/lesc/components/ChairAvailabilityCard.jsx
+++ b/client/src/lesc/components/ChairAvailabilityCard.jsx
@@ -4,7 +4,7 @@ import { inflect } from 'inflection';
 
 function ChairAvailabilityCard ({
   availableChairs,
-  reservedCount,
+  heldCount,
   occupiedCount,
   actionLabel,
   onActionClick,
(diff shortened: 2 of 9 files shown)


saadabdali:
One more quick thing:
- The PR also includes a new doc to document the state machine for deflections, which I think is not succinctly described anywhere else (`docs/holds-state-machine.md`). And added pointers to this file from the `ARCHITECTURE.md` and a couple relevant places in the code.

PR #915, opened 2026-05-06
fix icon routes

aaronhans:
The route for the icon referenced in the source HTML is returning HTML instead of the icon png file. This fix allows the images to be delivered


saadabdali (approved):
Looks good to me.

PR #916, opened 2026-05-06
Show warning banner in non-prod environments

### LINKED ISSUE #760: Make it easy for users to hit Prod rather than Dev
- In the past, we have had many users test the app on a Dev URL (e.g. reset.careconnectsf.org)
- Now we have a different production URL (I think reset.careconnect.sf.gov)
- This has caused a few flavors of problem:
  - Some users bookmarked and logged into the dev app, which caused confusion and problems (e.g. their data appeared to be out of sync with others in the room, because some users were pointing to a different instance of CareConnect!)
  - We handled this by shutting down all the accounts on the dev instance. But this creates a different user confusion (why isn't my login working anymore? and also how do I even get the new URL?)

**Suggestion:**
- We use (or create) some per-deployment env var that tracks whether this is a production deployment or not
- On any non-production deployment, we display a prominent banner at the top of the app that makes it obvious you're on a dev/test instance, and **includes a prominent link** to the real Prod instance

Alternatively, we could automatically redirect users. But that could be confusing. And surely there are *some* situations where we want to be able to use dev (or even want to allow others to use dev).

saadabdali:
Closes #760 , #912 

## In this PR

The problem this tries to solve is: "real users keep accidentally hitting the staging URL and being confused about why the data is not there, or why they can no longer log in. let's make it extremely obvious that they're not on the right URL, and easy to go to the right URL."

- We check for an environment variable `VITE_ENVIRONMENT_LABEL=PROD`, which tells us that our current environment is production.
- If that exists, do nothing.
- If it doesn't exist, display a prominent, un-dismissable banner on every page that says `This is a test site` and provides a link to production CareConnect.
- If `VITE_PRODUCTION_URL_OVERRIDE` is set, the link will use the URL provided there. Otherwise it falls back to the primary Production URL `https://reset.careconnect.sf.gov/login`, hard-coded.

Note that the env vars are **build-time** for Vite — they must be set when the deploy runs `npm run build`, not at server runtime.

### Deployer instructions (Production)
Set `VITE_ENVIRONMENT_LABEL=PROD`. This is the only way to hide the banner. Any other value, or omitting it, shows the banner.

### Deployer instructions (Dev/Staging/Etc)
Leave `VITE_ENVIRONMENT_LABEL` unset, or set it to anything except `PROD`

### Developer instructions

If you do nothing, then you will see the yellow banner in your local development and testing.

**You may wish** to hide the banner locally, to get a more accurate view of what a real user would see.) To do this:
1. In your local `client/.env`, add: `VITE_ENVIRONMENT_LABEL=PROD`.
2. Restart your docker container to pick up the env change.

## Screenshots

IMG_0240
IMG_0239
IMG_0238
IMG_0237
IMG_0236
IMG_0235

## Caveats
- The banner appears on almost every page in the app. I deliberately left out full-screen modals and the full-page camera views. Didn't seem necessary to mess with the layout on those pages.
- Right now this designed as an extremely binary feature. It crudely differentiates between "prod" and "non-prod", and is focused on **minimizing confusion for laypeople who accidentally hit a non-prod page**. It doesn't show fine distinctions between environments.
- I could have called the flag `HIDE_ENVIRONMENT_BANNER` (with `true`/`false`) instead of asking for a `VITE_ENVIRONMENT_LABEL`. But in the future it's possible we'd wire additional functionality that's environment-sensitive, or want to respond in a more granular way to the different environments, or make use of the provided label. So I did it this way.

### DIFF
--- a/client/src/components/Header.jsx
+++ b/client/src/components/Header.jsx
@@ -1,11 +1,21 @@
 import { Box, Container } from '@mantine/core';
 import classNames from 'classnames';
 
+import { useStaticContext } from '@/StaticContext';
+import { ENVIRONMENT_BANNER_HEIGHT, shouldShowEnvironmentBanner } from '@/components/EnvironmentBanner';
+
 import classes from './Header.module.css';
 
+// This is the page-level "fixed at viewport top" header that some detail/form
+// pages use as their own custom header (with a back button, etc.). It would
+// otherwise overlay the EnvironmentBanner because both want to live at top: 0.
+// We offset by the banner's height when the banner is showing so the banner
+// stays visible above this header.
 function Header ({ children, className }) {
+  const { env } = useStaticContext();
+  const top = shouldShowEnvironmentBanner(env) ? ENVIRONMENT_BANNER_HEIGHT : 0;
   return (
-    <Box className={classNames(classes.header, className)}>
+    <Box className={classNames(classes.header, className)} style={{ top }}>
       <Container className={classes.headerContainer}>
         {children}
       </Container>
--- a/client/src/AppLayout.jsx
+++ b/client/src/AppLayout.jsx
@@ -1,5 +1,5 @@
 import { useEffect } from 'react';
-import { AppShell } from '@mantine/core';
+import { AppShell, Box } from '@mantine/core';
 import { useDisclosure } from '@mantine/hooks';
 import { matchPath, useLocation, useNavigate } from 'react-router';
 import { useQueryClient } from '@tanstack/react-query';
@@ -8,12 +8,20 @@ import Header from './Header';
 
 import Api from './Api';
 import AppRoutes from './AppRoutes';
+import EnvironmentBanner, {
+  ENVIRONMENT_BANNER_HEIGHT,
+  shouldShowEnvironmentBanner,
+} from './components/EnvironmentBanner';
 import { useFacilityContext } from './FacilityContext';
+import { useStaticContext } from './StaticContext';
+
+const HEADER_HEIGHT = 80;
 
 function AppLayout () {
   const [opened, { close, toggle }] = useDisclosure();
   const navigate = useNavigate();
   const { facility, setFacility } = useFacilityContext();
+  const { env } = useStaticContext();
   const queryClient = useQueryClient();
 
   async function logout (event) {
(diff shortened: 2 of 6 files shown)


grahamplace:
> You may wish to hide the banner locally, to get a more accurate view of what a real user would see.) 
> In your local client/.env, add: VITE_ENVIRONMENT_LABEL=PROD.

Personally, I would _never_ want this when running a local dev server, I obviously know it's not prod since I'm actively developing on localhost, and I agree that for dev you want the UI to be as real as possible for layouts/sizes/etc. Setting `VITE_ENVIRONMENT_LABEL=PROD` in my dev env feels wrong, and could cause future confusion if anyone adds anything else that uses `VITE_ENVIRONMENT_LABEL`

What if instead the banner can _only_ render if `import.meta.env.PROD`, i.e. a production _build_ in an environment besides the live prod env? (since that's what we're actually targeting with this banner addition, not developers' local dev servers)

client/src/components/Header.jsx
```diff
@@ -1,11 +1,21 @@
...
+import { ENVIRONMENT_BANNER_HEIGHT, shouldShowEnvironmentBanner } from '@/components/EnvironmentBanner';
+
 import classes from './Header.module.css';
 
+// This is the page-level "fixed at viewport top" header that some detail/form
+// pages use as their own custom header (with a back button, etc.). It would
+// otherwise overlay the EnvironmentBanner because both want to live at top: 0.
+// We offset by the banner's height when the banner is showing so the banner
+// stays visible above this header.
 function Header ({ children, className }) {
```
grahamplace, line 14: [Devin review](https://app.devin.ai/review/sfbrigade/care-connect/pull/916) caught this, I confirmed it does cause this issue with the scan modal layout

before:
Image

after:

Image


Devin writeup:
```
The components/Header.jsx now applies a top offset based on shouldShowEnvironmentBanner, but this is incorrect when the Header is rendered inside a fullscreen Modal (e.g. ScanCodeModal at client/src/components/ScanCodeModal.jsx:150). Inside the modal, the environment banner is hidden behind the overlay, so the offset creates an unintended 40px gap.

Possible approaches:
1. Accept an optional prop (e.g. `disableBannerOffset`) on components/Header.jsx and pass it from ScanCodeModal.
2. Instead of computing the offset inside Header, let callers pass the top value explicitly (or default to 0) so context-aware parents can control it.
3. Use a React context or ref to detect whether the Header is inside a Modal and skip the offset.
```

PR #918, opened 2026-05-06
Shows person details after release (#914)

beaudrykock-moi:
## Summary

- Allow Care users to open the usual details view for post-release person records.
- Show the `Details` button on Care cards for `EXITED` records, including people who exited without legal release and people who were legally released then exited.

Closes #914


aaronhans (approved):
Cool, includes the specific states care staff would be concerned with also exited by design.

Doesn't include FAILED_INTAKE which should be a temporary state because they would hand that person over to on staff law enforcement to arrange transport to next destination, jail or hospital.

PR #925, opened 2026-05-07
Adds release narrative at SFSO request (#919)

beaudrykock-moi:
## Summary

- Appends the required “can care for themselves” medical-staff paragraph to the 849(b) release narrative during SOBERED custody release _only_.
- Persists the appended narrative on `Deflection.releaseNarrative` before release forms are queued, generated, stored, or e-mailed.
- Uses release confirmation date/time, released subject first/last name, and releasing deputy first initial, last name, and star number.

Closes #906

### DIFF
--- a/server/routes/api/deflections/release.js
+++ b/server/routes/api/deflections/release.js
@@ -139,9 +139,7 @@ export default async function (fastify, opts) {
           }
 
           const now = new Date();
-          const releasingDeputy = await tx.user.findUnique({
-            where: { id: request.user.id },
-          });
+          const releasingDeputy = request.user;
           const previousSubjectStatus = deflection.subjectStatus;
           // `sobered` releases from a pre-chair hold finalize immediately as
           // EXITED. Medical, behavioral-health, and "other" releases also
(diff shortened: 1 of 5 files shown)


server/routes/api/deflections/release.js
```diff
@@ -127,13 +128,18 @@ export default async function (fastify, opts) {
...
+              subject: true,
+            },
           });
 
           if (!RELEASABLE_STATUSES.includes(deflection.subjectStatus)) {
             throw conflictError(`Deflection ${id} cannot be released: status is ${deflection.subjectStatus}, expected one of [${RELEASABLE_STATUSES.join(', ')}]`);
           }
 
           const now = new Date();
+          const releasingDeputy = request.user;
```
saadabdali, line 142: This was originally a new transaction/query. But it looked unnecessary since I think request.user is already fully-hydrated with the fields we need, so I simplified here. Tested the generated 849b appended text before and after my change, and it looked identical.

saadabdali (approved):
I made the following changes:

- `release.js` seemed to be unnecessarily making a new query to retrieve the current user's info. It looked like we already had the relevant info without the query, so I removed it.
- The PR originally introduced a few name-formatting functions (to produce names like `J Doe` and `John Doe` to use in the appended text). The app already had some existing functions for that, so I  modified the PR to re-use the existing functions rather than introducing new ones.

PR #926, opened 2026-05-07
Upper casing guards on org unit

beaudrykock-moi:
## Summary

- Uppercase unit names anywhere users or admins type/select units.
- Normalize unit names before persistence when new units are created through user profile/unit selection flows or admin unit management.
- Force unit display to uppercase across profile, header, managed-user, and admin unit screens.
- Uppercase unit values before inserting them into legal PDF forms.
- Add a Prisma migration to retrospectively uppercase and trim existing `Unit.name` values.

Closes #919


beaudrykock-moi:
@aaronhans I removed the migration for this; would appreciate your review for potential deploy today

francisli (approved):

PR #931, opened 2026-05-07
all links generate pdf, open as downloadable

### LINKED ISSUE #921: User tries to view 849b PDF but doesn't see it/can't find
Repro:

1. Custody user was logged in with app in PWA mode on iOS/Safari
2. They released someone
3. They went to tap "849(b).pdf" button from person details
4. App loaded a new view that was blank, with an AWS url in the browser bar
5. User could not find any downloaded file, including in iOS Files app

aaronhans:
## Problem

PWA links were opening in no browser chrome mode and sometimes displaying PDF or downloading file in background and then user was stuck on that page and didn't know how to get back.

## Solution

All pdf links now:

- Hit API endpoint to generate pdf
- Gets file object and downloads it
- Shows toast to user to explain what is happening

This standardizes all pdf activity so it behaves the same all the time and works reliably in the PWA experience, doesn't take over the window sometimes. Users will have to learn where the files go on their device but toasts explain it and this behavior is universal so they should get familiar with it.

## Note

When I tried to preserve the retrieval of pregenerated pdf urls I we run into different behavior and couldn't reliably preserve the download behavior so with @beaudrykock-moi's approval switched to regen the pdf always.

Close #921

### DIFF
--- a/client/src/lesc/components/Deflection.jsx
+++ b/client/src/lesc/components/Deflection.jsx
@@ -123,7 +123,7 @@ function Deflection () {
     openInBrowser(url, `647f-${deflection.id}.pdf`)
       .then(() => {
         removeToast(toastId);
-        showToast('647(f) form ready', 'success', 4000, 'Open it from your downloads to view or print.');
+        showToast('647(f) form ready', 'success', 4000, 'Open your downloads/Files app to view or print.');
       })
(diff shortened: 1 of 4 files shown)


beaudrykock-moi (approved):
Looks good to me, no notes.

PR #932, opened 2026-05-07
Removes sfso incident email (#928)

beaudrykock-moi:
## Summary

Removes Sfso-incidentreports@sfgov.org email from any generated emails, per SFSO request.

Closes #928


aaronhans (approved):
clean removal

PR #935, opened 2026-05-07
Fix/886 rename legal release option

beaudrykock-moi:
## Summary

This PR clarifies custody release/exit flows so the UI matches the actual behavior for people who are not following the standard in-chair legal release path.

#### Changes to pre-`IN_CHAIR` person statuses release/exit flows 
- In overflow menu, renames "Legal release" to "Record exit to other": legal release in the standard "can care for themselves" flow should not be an option for these statuses. 
- Routes this new “Record exit to other” option through the existing release & exit flow with `Other` preselected and supplemental reason/destination fields visible.
- Removes “Can care for themselves” as a reason, since this should not be an option

#### Updates to `IN_CHAIR` happy-path “Start legal release” flow (main CTA)
  - Hides release reason chips (because "can care for themselves" is the default)
  - Shows fixed copy: “Release reason: can care for themselves”
  - Adds helper text directing users back to overflow actions for jail, hospital, or other exits

#### Updates to immediate release-and-exit flows:
  - Title changed to: “Confirm legal release and exit”
  - Helper text changed to: “When you confirm release and exit, the 849(b) will be sent to SFSO records and your e-mail.”

#### Updates to failed-intake behavior
  - Person card CTA: “Release and exit”
  - Detail primary CTA: “Release and exit”
  - Overflow menu only shows “Exit to jail”
  - Legal release view uses release-and-exit title/helper copy

Closes #886

### DIFF
--- a/client/src/lesc/components/custody/CustodyCard.jsx
+++ b/client/src/lesc/components/custody/CustodyCard.jsx
@@ -139,7 +139,7 @@ function CustodyCard ({ deflection, highlighted }) {
                 size='md'
                 onClick={() => navigate(`/custody/${deflection.id}/legal-release`)}
               >
-                Legal release
+                Release and exit
               </Button>
             )}
             {showStartRelease && (
--- a/client/src/lesc/components/custody/CustodyCard.test.jsx
+++ b/client/src/lesc/components/custody/CustodyCard.test.jsx
@@ -104,12 +104,13 @@ describe('CustodyCard', () => {
     expect(screen.queryByText(/Transfer code:/)).not.toBeInTheDocument();
   });
 
-  it('shows intake-not-completed banner and legal release for failed intake', () => {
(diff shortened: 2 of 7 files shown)


saadabdali:
Taking a look here.

saadabdali (approved):
Code looks okay. Few notes:

- In this PR, if a subject reaches the `FAILED_INTAKE` state, the only options are `Release & Exit` and `Exit to jail`
- That means there is no `Record death` option at all. If a death occurs here, the app offers no way to capture it.
- It also seems _slightly_ odd that we no longer show the `Exit to hospital` option in the overflow (`...`) menu in this situation, even though we do still display it in other situations.

If this is intentional, we can proceed. Otherwise we may want to rework these.

More generally I wonder if we want to create some intermediate artifact (a diagram of all the subject states, with arrows between them showing the state transitions we intend to allow)? Right now that information isn't really captured anywhere other than in the code itself.

PR #944, opened 2026-05-09
Updated subject and body with correct statement

justin123yu:


francisli (changes requested):
@justin123yu you modified the `html.ejs` version of the transfer-form email body, but not the `text.ejs` template. They should have matching content (minus HTML formatting).

You modified the test to verify the subject changes. Can you also add a check to verify the body text is as expected?

justin123yu (commented):
Made Changes stated by Francis

francisli (approved):

PR #951, opened 2026-05-11
Allows custody user to edit property (#893)

beaudrykock-moi:
## Summary

- Adds custody access to edit personal property details after custody transfer and before exit.
- Reuses the existing property edit form at a new `/custody/:id/property` route.
- Shows success/error toasts after custody property saves and returns users to the custody details overview.
- Tightens backend property permissions so custody property fields/photos follow the same after-transfer-before-exit window.

Closes #893

### DIFF
--- a/server/routes/api/property-photos/delete.js
+++ b/server/routes/api/property-photos/delete.js
@@ -2,6 +2,7 @@ import { StatusCodes } from 'http-status-codes';
 import { z } from 'zod';
 
 import PropertyPhoto from '#models/propertyPhoto.js';
+import { canModifyDeflectionProperty } from '#lib/incidentPermissions.js';
 
 export default async function (fastify, opts) {
   fastify.delete('/:id',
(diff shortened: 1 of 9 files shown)


server/routes/api/property-photos/delete.js
```diff
@@ -22,13 +23,14 @@ export default async function (fastify, opts) {
...
         where: { id },
+        include: { deflection: true },
       });
 
       if (!data) {
         return reply.code(StatusCodes.NOT_FOUND).send();
       }
 
-      if (data.createdById !== request.user.id && !request.user.isAdmin) {
+      if (!canModifyDeflectionProperty(data.deflection, request.user)) {
```
aaronhans, line 33: This change in logic does have some implications. Previously the delivering SFPD officer could pass this check and now they don't because it looks for currentOfficerId (which becomes null after custody transfer). Confirmed with @beaudrykock-moi that this is desired, we want to limit editing after handoff and UI doesn't support that anyway.

aaronhans (approved):
Confirmed functionality locally, SFSO can edit personal property details, no code issues

PR #959, opened 2026-05-14
Baked appearances and fonts to ensure fields print on mobile (#958)

beaudrykock-moi:
## Summary

This PR makes generated `849b` PDFs print-safe on mobile by baking field appearance streams into the PDF at generation time instead of relying on the viewer to render them.

### Details

- Updated `server/lib/forms/849b/fill849b.js` to:
  - embed `StandardFonts.Helvetica`
  - call `form.updateFieldAppearances(appearanceFont)`
  - remove `AcroForm.NeedAppearances`
  - save the PDF normally

- Added a regression assertion in `server/test/lib/forms/849b.test.js` to verify generated `849b` PDFs no longer depend on `NeedAppearances`.

### Why

The previous implementation filled form fields but saved with viewer-side appearance generation:

- `NeedAppearances = true`
- `save({ updateFieldAppearances: false })`

That can display correctly in desktop PDF viewers while still printing blank from mobile Outlook / mobile print pipelines, because the file depends on the client viewer to synthesize printable field appearances.

This change makes the generated PDF self-contained and more portable across mobile viewers and print paths.

Closes #958

### DIFF
--- a/server/lib/forms/849b/fill849b.js
+++ b/server/lib/forms/849b/fill849b.js
@@ -10,7 +10,7 @@
  *   const filledBytes = await fill849b(templatePdfBytes, data);
  */
 
-import { PDFDocument, PDFName, PDFBool } from 'pdf-lib';
+import { PDFDocument, PDFName, StandardFonts } from 'pdf-lib';
 
 // ═══════════════════════════════════════════════════════════════════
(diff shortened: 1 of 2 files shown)


aaronhans (approved):
I don't see any issues here, everything works the same for me. This doesn't fix the issue everywhere so we should continue to explore optimizations but improving the issue in Outlook on iOS is helpful

PR #981, opened 2026-07-08
need to restrict API to SFPD arrests only per Sheriff

aaronhans:
Sheriff clarified that the API provided to SFPD should not include any arrests they did not initiate. Most arrests will be initiated by SFPD but Sheriff could initiate arrests as well.


beaudrykock-moi (approved):
Looks good to me! Clearly and simply excludes non-`sfpd` orgs, including any `null` values for `createdByOrganizationId`.

PR #985, opened 2026-07-23
[Closes #980] Extend visibility of exited records to 72 hours

grahamplace:
## Summary

Extends visibility of exited patient records from 24 to 72 hours after exit, so staff can use the data for auditing (#980).

- Adds `server/lib/retention.js` with a shared `EXITED_VISIBILITY_HOURS = 72` constant and `exitedVisibilityCutoff()` helper, replacing the two inline `24 * 60 * 60 * 1000` literals in the deflections list endpoint. The window is still measured from `exitedAt`.
- Guards the anonymization job against the >24h-stay edge case: with a 72h window, a long stay pushes `exit + 72h` past the 96h-from-creation redaction backstop, which could have redacted PII (and hidden the record) while it was still supposed to be visible. `subject.anonymize()` now defers redaction until every deflection's exit is older than 72h — effectively `max(createdAt + 96h, latestExit + 72h)`. Never-exited subjects still redact at the 96h backstop, so stuck records can't retain PII indefinitely.
- Updates the custody empty-state copy (and its story) from 24 to 72 hours.

## Testing

- New `server/test/routes/api/deflections-visibility.test.js` pins the 72h boundary (71h visible / 73h hidden) across all three list branches: `scope=history`, `subjectStatus=EXITED`, and mixed `subjectStatus=RELEASED,EXITED`. Verified these fail when the window is set back to 24h.
- Three new subtests in `anonymizeSubjects.test.js`: defers redaction at exit+71h, redacts at exit+73h, and the most recent exit governs subjects with multiple deflections.
- Full server suite passes (610 tests); lint clean on changed files.

## Note for reviewers

The `ANONYMIZE_CUTOFF_HOURS` env override still controls the creation-time backstop, but the new exit guard can defer redaction past it for long stays. If any environment sets this variable with different expectations, worth confirming before merge.

### DIFF
--- a/server/lib/retention.js
+++ b/server/lib/retention.js
@@ -1,9 +1,5 @@
-// PII retention windows.
-//
-// EXITED deflections remain visible in list results for this long after exit
-// so staff can use the data for auditing (issue #980). The anonymization job
-// (prisma/client.js) also waits for this window to close before redacting a
-// subject with an exited deflection, so records never disappear mid-window.
+// How long EXITED deflections stay visible in list results, measured from
+// exitedAt. The anonymization job also waits for this window to close.
 export const EXITED_VISIBILITY_HOURS = 72;
 
 export function exitedVisibilityCutoff (now = new Date()) {
--- a/server/lib/retention.js
+++ b/server/lib/retention.js
@@ -1,0 +1,11 @@
+// PII retention windows.
+//
+// EXITED deflections remain visible in list results for this long after exit
+// so staff can use the data for auditing (issue #980). The anonymization job
+// (prisma/client.js) also waits for this window to close before redacting a
+// subject with an exited deflection, so records never disappear mid-window.
+export const EXITED_VISIBILITY_HOURS = 72;
+
+export function exitedVisibilityCutoff (now = new Date()) {
+  return new Date(now.getTime() - EXITED_VISIBILITY_HOURS * 60 * 60 * 1000);
+}
--- a/server/prisma/client.js
+++ b/server/prisma/client.js
@@ -137,9 +137,8 @@ const prisma = new PrismaClient({
         const parsed = parseFloat(process.env.ANONYMIZE_CUTOFF_HOURS);
         const hours = Number.isFinite(parsed) ? parsed : 96;
         const cutoff = new Date(now.getTime() - hours * 60 * 60 * 1000);
-        // Stays longer than 24h push exit + 72h past the creation backstop, so
-        // also hold anonymization until every exit's visibility window (issue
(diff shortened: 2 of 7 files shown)


server/lib/retention.js
```diff
@@ -0,0 +1,7 @@
+// How long EXITED deflections stay visible in list results, measured from
+// exitedAt. The anonymization job also waits for this window to close.
+export const EXITED_VISIBILITY_HOURS = 72;
+
+export function exitedVisibilityCutoff (now = new Date()) {
```
grahamplace, line 5: We previously did `const twentyFourHoursAgo = new Date(Date.now() - 24 * 60 * 60 * 1000);` twice in list.js to do stuff related to the 24hr exit visibility window. Prefer to unify into a helper + constant here so various use cases of post-exit visibility computation stay in sync

server/prisma/client.js
```diff
@@ -132,15 +133,24 @@ const prisma = new PrismaClient({
     },
     subject: {
       async anonymize (now = new Date()) {
```
grahamplace, line 135: We previously just always anonymized data 96hrs after creation to keep this simple. That assumed exit always happens within 24h of the record being created — but the record is created at detention, so extended holds, transport time, and reopened records can push exit later, and with the new 72hr post-exit window the 96hr anonymization could blow away data that's still supposed to be visible.

So here we're saying: don't anonymize anyone with a deflection that exited less than 72hrs ago. Records that never exited still anonymize at the 96hr backstop so stuck records can't hold PII forever

server/routes/api/deflections/list.js
```diff
@@ -63,11 +64,10 @@ export default async function (fastify) {
           { status: Deflection.HoldStatus.ACTIVE },
           { subject: { isNot: null } },
         ]);
-        // Preserve the existing 24-hour cap on EXITED rows.
-        const twentyFourHoursAgo = new Date(Date.now() - 24 * 60 * 60 * 1000);
+        // Cap visibility of EXITED rows to the retention window.
         addOrGroup([
           { subjectStatus: { not: Deflection.SubjectStatus.EXITED } },
-          { exitedAt: { gte: twentyFourHoursAgo } },
+          { exitedAt: { gte: exitedVisibilityCutoff() } },
```
grahamplace, line 70: This is the part of the PR that directly solves https://github.com/sfbrigade/care-connect/issues/980 — we want to show deflections for 72hrs post-exit instead of 24, so we just change this filter's window to use 72hrs

beaudrykock-moi (approved):
Looks exactly as required to me - thank you! (Reviewed for sense not code quality, not being qualified for the latter :) )

francisli (approved):

PR #988, opened 2026-08-03
No chromium

aaronhans:
There were a couple things actually using chromium: 1 remaining pdf and mermaid diagram creator. Browser engines could also be installed for playwright and we aren't currently omitting npm dev package installs from our image but we don't install the playwright browser engines there so we didn't have to touch that to get rid of chromium.

1. The last form using browser rendering: Narcotics release
- Remove chromium by creating the narcotics release (possible 2nd page of release certificate) via pdf-lib instead
- fix how numbers at top of 849b are created so they aren't huge (SFSO complaining about this)
- dead file removal after last chromium built pdf removed:

| File | Why it's safe to delete |
|---|---|
| `shared/renderReactForm.js` | The only puppeteer-core/react-dom entry point; nothing imports it |
| `shared/FormContainer.jsx` | React document shell; only consumed by the dead render path |
| `shared/formComponents.jsx` | React primitives; only used by the dead JSX |
| `shared/pdf-forms.css` | Imported only by `FormContainer.jsx` |
| `cert/NarcoticsNotice.jsx` | Replaced by `cert/narcoticsNotice.js` (pdf-lib) |
| `647f/Form647f.jsx` | Legacy React version of the **already-live fillable 647f**; not imported anywhere |
| `esbuild.forms.js` | Only compiled the two dead JSX entry points |
| `dist/.gitignore` | The dir exists only to hold esbuild output |

2. The mermaid diagram creation
- puppeteer is a dependency of mermaid cli under prisma erd generator
- the live regeneration of the mermaid database diagram is important but we are only currently creating it by writing markdown. The viewable diagram in the github repo is an image because github interprets the mermaid markdown syntax as an image. We never used the puppeteer based actual image file generation
- therefore we can rip out puppeteer from this system and not lose our current dynamic db image
- unfortunately we had to stub mermaid cli to get rid of the puppeteer dependency
- In the future we could remove all dev dependencies from the package but there are additional issues with that we need to unravel so that wasn't attempted here.

```
prisma-erd-generator                     ← KEPT — ERD.md keeps auto-generating
└─ @mermaid-js/mermaid-cli   → STUB      (~55M gone)
   ├─ mermaid, chalk, ...    → gone
   └─ puppeteer (peer, ^23)  → gone      ← the Chromium/CVE carrier
      ├─ puppeteer-core@23   → gone
      └─ @puppeteer/browsers → gone
```


beaudrykock-moi (approved):
Looks good to me - appropriate and well scoped.

PR #997, opened 2026-08-13
Bugfix: Org admins with a Unit get a 500 on user GET/PATCH

### LINKED ISSUE #992: Org admins with a unit assigned get a 500 on user GET/PATCH
## Summary

Org admins who **have a unit assigned** get a 500 when they view or edit another user in their organization. `GET /api/users/:id` and `PATCH /api/users/:id` are both affected.

```
TypeError: Cannot read properties of undefined (reading 'includes')
  server/models/user.js:119   get isOrgAdmin -> return this.roles.includes('ORG_ADMIN');
  server/models/base.js:40    Object.get     -> return get.call(receiver);
  server/routes/api/users/get.js:41          -> if (!requestUser.isOrgAdmin || ...)
```

## Who is affected

- **Broken:** org admin *with a unit assigned*, acting on another user in their org (GET and PATCH).
- **Not affected:** org admins with no unit; platform admins (`isAdmin` short-circuits earlier); users acting on their own record.

The unit is the trigger, not incidental — see below.

## Root cause

`server/routes/api/users/get.js:40` and `server/routes/api/users/patch.js:121` do:

```js
const requestUser = new User(request.user);
```

But `request.user` is **already** a `User` model — `server/plugins/auth.js:39` constructs it. Re-wrapping an existing model silently discards its data:

1. `User`'s constructor (`server/models/user.js:94`) sees `data?.unit?.name` and rebuilds data as `{ ...data, unit: {...} }`.
2. `data` here is the Proxy returned by `Base` (`server/models/base.js:26`). That proxy defines only `get` and `set` traps — no `ownKeys` — so spreading it copies only the wrapped object's own keys: `changes`, `__fields`, `__data`, `update`. Every real field (`roles`, `id`, `organizationId`, ...) lives in a closure and is dropped.
3. The re-wrapped model's `get` trap sees `roles` in `Prisma.UserScalarFieldEnum` and returns it from the gutted data object → `undefined`.
4. `isOrgAdmin` (`user.js:119`) is the only role getter without optional chaining, so it throws rather than returning `false`.

A user with no unit skips step 1 entirely, so the redundant wrap forwards correctly and nothing breaks. That is why this got through review and why it only surfaced in production.

Introduced in 65f0424e (PR #822, "[Closes #762] Adds org admin ability to edit person details"). Those two lines are the only occurrences of `new User(request.user)` in the server; every other call site reads getters directly off `request.user` (e.g. `deflections/list.js:122`, `deflections/cancel.js:51`, `plugins/auth.js:66`).

## Proposed fix

1. **Remove the redundant re-wrap** at `get.js:40` and `patch.js:121`; use `request.user.isOrgAdmin` directly, matching the rest of the codebase. This is what stops the 500s.
2. **Align `isOrgAdmin` with its sibling getters** — `this.roles?.includes(User.Role.ORG_ADMIN) ?? false`. Note this must ship *with* fix 1, not instead of it: on its own it converts the crash into a silent, wrong `403` for the same users.
3. **Make the constructor double-wrap-safe** as insurance, since `new User(someUser)` looks harmless and mostly works:
   ```js
   if (data instanceof User) {
     data = data.toJSON();
   }
   ```
   Verified against the real model: `instanceof` works through the proxy (`Base` defines no `getPrototypeOf` trap), `toJSON()` returns the complete data, and plain Prisma rows return `false` so normal construction is untouched.
4. **Regression test:** org admin *with a unit* fetching/patching another user in their org. Without the unit assigned, the test passes even with the bug present.

## Explicit non-goal

Adding `ownKeys`/`getOwnPropertyDescriptor` traps to `Base` so that spreading a model works in general. That is arguably the deepest fix, but `Base` backs every model in the app, proxy invariants around `ownKeys` are fussy, and it would change the behavior of every `{ ...model }` / `Object.keys(model)` path in the codebase. That deserves its own PR and test pass, not a ride-along with a production 500 fix.

saadabdali:
Closes #992 

## Repro

- Log in as a user that has Org Admin privilege and has a Unit (in my local dev setup, `sfso2@` meets these criteria)
- From the menu, choose `Manage Users`
- Click on any user that's not myself
**- Before this PR:**
  - Page fails to render
  - Server throws a 500 due to `Cannot read properties of undefined`
- **With this PR:**
  - page loads, no server error
  - able to view, edit, re-save user details for a managed user

The issue is that we were taking an existing `User` object and passing it back into the constructor for `User`, which didn't throw an immediate error, but caused problems downstream because some object properties get dropped (in this case `roles`). I'll comment directly on the code diff for more detail.


francisli (approved):
Thanks, I encountered this on Monday and looked briefly into it but didn't have the time to chase it down!

PR #1005, opened 2026-08-13
Add double-confirmation flow for failing intake

### LINKED ISSUE #995: As Intake Staff, I can double-confirm I want to reject someone in medical intake
So that mistakes don't lead to losing the person's record from CareConnect.

Design notes
This is currently implemented as a destructive action without double confirmation. On a couple of occasions, users have accidentally rejected someone at intake. Rather than build a complex undo functionality, let's add additional confirmation so that this destructive action requires more thought.

AC:
- [ ] When an IS user updates status and taps "Intake failed - return to Deputy", this triggers a secondary confirmation

saadabdali:
Resolves #995 

Based on the changes specified in [this design](https://www.figma.com/design/Q8kS4FJXh1TbbM8eDR7Z6Y/CareConnect?node-id=19420-36858&t=iGd0U9kNSXnZ7zOD-0). 

https://github.com/user-attachments/assets/25160450-5349-4afc-a225-921517d881b8

### DIFF
--- a/client/src/lesc/components/care/CompleteIntakeModal.jsx
+++ b/client/src/lesc/components/care/CompleteIntakeModal.jsx
@@ -52,7 +52,7 @@ function CompleteIntakeModal ({
         ? (
           <Stack gap='2xl'>
             <Stack gap='sm'>
-              <Group justify='space-between' align='flex-start' wrap='nowrap'>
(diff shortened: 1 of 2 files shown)


beaudrykock-moi (approved):
Looks good to me.

